PR #6, opened 2016-06-07
PIP issue resolved, Log folder creation & Manifest update

KohliRocks:
PIP issue resolved 
Log folder creation in python code
Manifest update

### DIFF
--- a/liota/__init__.py
+++ b/liota/__init__.py
@@ -35,7 +35,7 @@ import json
 import logging
 import logging.config
 import os
-import time
+import errno
 import ConfigParser
 from utilities.utility import systemUUID, findLiotaConfigFullPath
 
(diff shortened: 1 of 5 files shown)


GregBollella:
Looks good.
G

liota/__init__.py
```diff
@@ -45,6 +45,7 @@ def setup_logging(default_level=logging.WARNING):
 
     """
     log = logging.getLogger(__name__)
+    mkdir_log("/var/log/liota")
```
pmasrani: Can we read the log path from liota.conf instead of hard-coding here?

KohliRocks: Have, configured this path in liota.conf
Thanks

PR #8, opened 2016-06-18
Pint library added in liota package to support SI units and updated README.md

KohliRocks:


GregBollella:
Looks ok as far as it goes. I'd like to add more complex examples of using unit typing to avoid mistakes. Perhaps this is something our new intern can do as an initiation to liota?

I'd say go ahead and push this back to the mainline. 

G

PR #9, opened 2016-06-28
Optimized vROPs DCC

pmasrani:
1) Removed unnecessary sleep call in vROps registration.
2) Changed addition of properties to a single call, instead of one call
for each property.

### DIFF
--- a/liota/dcc/vrops.py
+++ b/liota/dcc/vrops.py
@@ -153,7 +153,7 @@ class Vrops(DataCenterComponent):
          }
       }
 
-    def properties(self, msg_id, res_uuid, res_kind, timestamp, properties_list):
+    def properties(self, msg_id, res_uuid, res_kind, timestamp, properties):
         msg = {
             "transationID": msg_id,
             "type": "add_properties",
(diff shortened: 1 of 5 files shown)


liota/dcc/vrops.py
```diff
@@ -165,12 +164,13 @@ def properties(self, msg_id, res_uuid, res_kind, timestamp, key, value):
                 "property_data": []
             }
         }
-        msg["body"]["property_data"].append({"propertyKey": key, "propertyValue": value})
+        for property_dict in properties_list: 
+            msg["body"]["property_data"].append({"propertyKey": property_dict.keys()[0], "propertyValue":  property_dict.values()[0]})
```
KohliRocks: This looks proper. Just one suggestion in case if we have multiple items in the dictionary as user will be providing this data(we don`t have control), the safe code to iterate over dictionary using the following call;

for key, value in property_dict.iteritems():

rather then just referring the first element.

KohliRocks:
Ship It !!!

winniex1:
ship it~

PR #10, opened 2016-07-06
Pint based example apps using simulated physical models

bfrggit:
This branch contains several **new files** in the `example` folder.

I. Two of them are **simulated physical models**:

- `bike_model_simulated.py`
- `thermistor_model_simulated.py`

An object of either classes will create a thread, that changes some states of the model randomly over time. There are getter methods that can be called from the app to retrieve the states.

II. Three **example applications using vROps DCC**:

- `vrops_bike_model_simulated.py`
- `vrops_thermistor_model_simulated.py`
- `vrops_thermistor_model_simulated_wrong.py`

The third one shows a wrong implementation which will cause an exception in runtime.

III. Three **example applications using graphite DCC**:

- `graphite_bike_model_simulated.py`
- `graphite_thermistor_model_simulated.py`
- `graphite_thermistor_model_simulated_wrong.py`

Similar to the vROps ones.

---
This branch also contains minor changes on other files.

I. `liota/things/function.py` (new file)

This file is almost the same to "ram.py". The only difference is the type string of the device, so our simulated devices appear as functions on vROps instead of RAMs.

II. `.gitignore`

Added two lines for two possible temporary files, "*~" and "nohup.out".


GregBollella:
Nice work Charles! I've seen the vrops graphs. Can you stop by sometime and let me see the graphs in Graphite?


Thanks, Greg

example/bike_model_simulated.py
```diff
@@ -0,0 +1,105 @@
+import threading
```
pmasrani: Can you add the Copyright?

example/bike_model_simulated.py
```diff
@@ -0,0 +1,105 @@
...
+def main():
+    bike = BikeModelSimulated()
+    bike.run()
+    while True:
+        print "slope:   {},\nrev:     {},\nload:    {}\n---".format(
+                bike.get_slope(),
+                bike.get_revolution(),
+                bike.get_weight_load()
+            )
+        time.sleep(5)
```
pmasrani: I believe you are trying to read a new set of bike properties every 5 seconds. Since both threads are running at the exact same frequency and are started together, in case of race conditions, if for one iteration the simulated thread gets read before reader thread and in the next iteration, the order gets reversed, chances are stale values will be read.

bfrggit: For the note regarding order of read/update - I don't think this is a big issue. And also, I don't think this is an issue you can solve on real physical objects. I cannot put a thread-safe lock on a real bike. 

What do you think?

KohliDev:
Ship It. A suggestion to add "bike_model_simulated.py,thermistor_model_simulated.py" in  a sub-folder under example folder.

PR #11, opened 2016-07-14
Remove call to connect_soc from constructor of TransportLayer.

d0ivanov:
Instead call that method in each derived class' constructor.


KohliDev:
Ship It !!! We will wait for one more review before merging it to the main branch. Thanks.

pmasrani:
Ship It!

PR #14, opened 2016-07-22
SI support for vROps DCC

bfrggit:
Included in this pull request are
1. `liota/dcc/dcc_base.py`
2. `liota/dcc/vrops.py`
3. `liota/utilities/si_unit.py`

Work done on those files
1. Added an abstract method `publish_unit` for the base DCC class.
2. Added implementation for this method in vROps. Units are sent to vROps using `set_properties`.
3. Added an `si_unit.py` in utilities, that contains necessary methods to parse unit into prefix and name strings. Testing codes are provided to demonstrate the parser works for units listed in SI standard documents, and the output names are consistent with those in SI documents. For derived units not listed in the documents, we deliver our best effort to parse them.
4. Improved formatting in `vrops.py` for two methods - they were indented with two spaces per level, and I changed that two four spaces.

Tested working in my VM.

Two units not supported by pint currently:
1. kat (You will have to use 'mol / s')
2. Sv (It is actually supported but it is implemented as an alias of Gy. Pint will always output its name as Gy for *gray* instead of Sv, though.)


liota/dcc/dcc_base.py
```diff
@@ -53,4 +53,8 @@ def subscribe(self):
         pass
 
     def create_metric(self, gw, details, unit, sampling_function, sampling_interval_sec=10, aggregation_size=6):
+        self.publish_unit(gw, details, unit)
```
pmasrani: should we make unit as an optional parameter, as not everything might have one (e.g., count of open file handles, etc) and passing null does not seem very intuitive? (Not actually related to your current changes.)

liota/utilities/si_unit.py
```diff
@@ -0,0 +1,324 @@
...
+
+    return ts
+
+#---------------------------------------------------------------------------
+# This is primary method that should be imported in vROps DCC, or any future
+# DCC that has or requires unit support, in order to parse unit parameter
+# passed through create_metric call.
+
+def parse_unit(unit):
+    pf = None
```
pmasrani: A lot of names in this CL seem to be non-intuitive (pf, un, ts etc.) and require a lot of effort to understand the meaning. Can you change the variable names?

liota/utilities/si_unit.py
```diff
@@ -0,0 +1,324 @@
...
+    ]
+unit_tables = lambda ureg: [
+        units_table_1(ureg),
+        units_table_2(ureg),
+        units_table_3(ureg),
+        units_table_4(ureg)
+    ]
+
+
+# Testing code
```
pmasrani: Generally, testing code is in a separate file/package/folder (which ideally do not even get installed along with the package installation from pip). Should we separate this out?

liota/utilities/si_unit.py
```diff
@@ -0,0 +1,324 @@
...
+def main():
+    ureg = pint.UnitRegistry()
+    print_split = "-" * 76
+
+    # Base units and examples of derived units defined in SI standard documents
+    def parse_unit_with_color(un):
+        pf = "\033[1;36mnull\033[0m"
```
pmasrani: These values are not intuitive, can you please add  a comment what do they mean?

liota/dcc/dcc_base.py
bfrggit: Maybe, but that would require a change to the interface because optional parameters cannot go before mandatory ones. I would say, that sounds reasonable but probably not a high priority one now.

liota/utilities/si_unit.py
bfrggit: Thanks. Will do this. Please see changes in next commit.

liota/utilities/si_unit.py
bfrggit: Do we have any place to put in testing code now?

liota/utilities/si_unit.py
bfrggit: Will improve this.

liota/utilities/si_unit.py
```diff
@@ -0,0 +1,238 @@
...
+    # Attempt to extract prefix using base unit conversion
+    tuple_base = ureg.get_base_units(unit)
+    if tuple_base[0] == 1: # Base unit, or simple combo of several base units
+        if tuple_base[1] != ureg.dimensionless:
+            obj_unit = unit
+    else: # Prefixed unit, or non-SI unit
+        try:
+            str_prefix = _get_prefix(tuple_base[0])
+            obj_unit = tuple_base[1]
+            # if re.compile("\s").search(str(obj_unit)) is not None:
```
pmasrani: commented code, remove.

test/si_unit.py
```diff
@@ -0,0 +1,147 @@
...
+        str_prefix, str_unit_name = parse_unit_with_color(obj_unit)
+        print "  " + (c_bold("%s") + " - %s, %s") % (
+                obj_unit,
+                str_prefix,
+                str_unit_name
+            )
+
+    print print_split
+
+# if __name__ == "__main__":
```
pmasrani: commented code, remove.

GregBollella:
Looks reasonable to me.

liota/dcc/vrops.py
```diff
@@ -155,26 +156,25 @@ def publish(self, metric):
         self.con.send(message)
 
     def init_relations(self, gw):
-      """ This function initializes all relations between gateway and it's children.
-          It is called after each object's UUID is received.
+        """ This function initializes all relations between gateway and it's children.
```
KohliDev: I am not sure if the spaces are added intentionally. Please use a python code editor to format the code.

KohliDev: Also please change the permission of utility.py back to original.

bfrggit: It was intentional. Because originally it was 2 spaces per level - only for these two methods, which was not consistent which any part of the same file or any file in the repo.

pmasrani:
Ship It!

PR #16, opened 2016-07-29
Metric should be able to receive callbacks that are bound methods written in python.

d0ivanov:
When you pass an instance method to create_metric, it is improperly called with two arguments since thce current check is whether or not the callback requires arguments or not. All instance methods in python require at least one parameter to be passed to them (the instance on which the method is called).


GregBollella:
Thanks for spotting this!

KohliDev:
Hello @d0ivanov ,

I think this pull request is out-dated due to the recent liota refactoring. Please raise a new pull request, if you still figure out that bug exist.

Thanks,
Kohli

PR #17, opened 2016-07-29
MQTT Support and basic example

KohliDev:
Basic MQTT support added to liota with QoS=1. An example has been added to showcase the use of this package to collect values in the respective queue from various devices (things) using MQTT protocol. In future, we also plan to add an example for publishing data from Gateway to DCC component using MQTT protocol. The code has been tested by creating dummy channels on test.mosquitto.org. 
Also, the usage of lambda function has been shown in order to pass parameters to user defined methods (UDM).
 Detailed comments are provided in the example file.

### DIFF
--- a/examples/dk300_edge_system_iotcc_graphite.py
+++ b/examples/dk300_edge_system_iotcc_graphite.py
@@ -205,6 +205,7 @@ if __name__ == '__main__':
     except RegistrationFailure:
         print "Registration to IOTCC failed"
 
+    # Multiple DCC support
     # Sending data to an alternate data center component (e.g. data lake for analytics)
     # Graphite is a data center component
     # Socket is the transport which the agent uses to connect to the graphite instance


liota/transports/mqtt.py
```diff
@@ -52,21 +54,30 @@ def __init__(self, url, port):
...
+        log.info("Message {0} sent on topic {1}".format(str(message), str(topic)))
 
     def subscribe(self, topic):
+        log.info("Subscribed to topic {0}".format(str(topic)))
         self.client.subscribe(topic, qos=1)
+        thread.start_new_thread(self.loop_thread, ())
+
+    def loop_thread(self):
         self.client.loop_forever()
+        pass
```
bfrggit: Why do we need a pass here?

liota/transports/mqtt.py
```diff
@@ -52,21 +54,30 @@ def __init__(self, url, port):
...
         self.client = paho.Client()
+        self.subscribe = self.subscribe
+        self.publish = self.publish
         self.client.on_message = self.on_message
         self.client.on_publish = self.on_publish
         self.client.on_subscribe = self.on_subscribe
         self.connect_soc()
         TransportLayer.__init__(self)
 
     def connect_soc(self):
```
bfrggit: Why is this method named `connect_soc`?

liota/transports/mqtt.py
```diff
@@ -52,21 +54,30 @@ def __init__(self, url, port):
         self.url = url
         self.port = port
         self.client = paho.Client()
+        self.subscribe = self.subscribe
+        self.publish = self.publish
```
bfrggit: It looks weird... What do these two lines do?

example/sampleProp.conf
```diff
@@ -1,6 +1,10 @@
 WebSocketUrl = "Webesocket-URL"
 vROpsUID = "Username"
 vROpsPass = "Password"
+MqttServer = "Mqtt-IP"
+MqttPort = "Mqtt-Port"
```
bfrggit: This kind of naming kind of suggests that we plan to use the same MQTT broker for device->gateway and for gateway->DCC communication. Is this our plan?

liota/transports/mqtt.py
GregBollella: Looking at line 76 and in the sample program it seems to me that the model is one thread per topic/metric. If so. then we need to re-architect this solution because of the max thread limitation. If not, then please point our where I am mistaken. 

Possibly the more scalable architecture is to have a single thread listening on all topics and pushing messages to the correct queue?


Greg

bfrggit: Looking at Ln 73-80 in the user app, I think we are going to subscribe on a big topic and use that `message_callback_add` to add call-backs to differentiate between sub-topics.

I haven't done this before, but it seems that `subscribe` will happen only once per user app.

example/vrops_mqtt.py
```diff
@@ -0,0 +1,122 @@
...
+    mqtt_conn = Mqtt(config['MqttServer'], config['MqttPort'])
+    # subscribe to the base mqtt channel example "SmartBuilding/#"
+    mqtt_conn.subscribe(config['MqttChannel'])
+    # add a message callback to the client of mqtt_conn object
+    # glass panel call back method added to the mqtt client to listen
+    # on specific sub-channel example "SmartBuilding/GlassPanel"
+    mqtt_conn.client.message_callback_add(config['MqttSubChannel1'], callback_glass_panel)
+    # solar lights call back method added to the mqtt client
+    # to listen on specific sub-channel example "SmartBuilding/SolarLights"
+    mqtt_conn.client.message_callback_add(config['MqttSubChannel2'], callback_solar_lights)
```
bfrggit: Ln 77 and Ln 80 operate on mqtt_conn.client directly from the user app, which does not look elegant.

example/sampleProp.conf
```diff
@@ -1,6 +1,10 @@
...
+MqttChannel = "Mqtt_channel"
+MqttSubChannel1 = "Mqtt_sub_channel_1"
```
winniex1: To coordinate your example code, MqttSubChannel2 is better to be listed also.

example/vrops_mqtt.py
winniex1: Agree. Could we provide a wrapper which just let user subscribe topic with queue? The MQtt transport can be in charge of dealing with mqtt.client, adding callbacks, dealing with received topic-based messages, and UDM can just read queue.

example/vrops_mqtt.py
```diff
@@ -0,0 +1,122 @@
...
+    else:
+        print "vROPS resource not registered successfully"
+
+    # Solar light device registered
+    solar_lights_device = Device(config['Device2Name'], 'Read', gateway)
+    vrops_solar_device = vrops.register(solar_lights_device)
+    if vrops_solar_device.registered:
+        # solar_lights queue passed as parameter to UDM get_value()
+        solar_metric = vrops.create_metric(vrops_solar_device, "Solar Lights", unit=None, sampling_interval_sec=0,
+                                            aggregation_size=1, sampling_function=lambda :get_value(solar_lights_queue))
```
winniex1: This is very useful! good example.

liota/transports/mqtt.py
GregBollella: Got it. If subscribe happens only once per user app maybe ok for now but what about when every metric potentially is a user app (considering the evolving package design). We have to make sure not to bind a thread to a metric or even to a device.

liota/transports/mqtt.py
winniex1: in old version, it's self.client.subscribe = self.subscribe, do you mean that?

PR #19, opened 2016-08-05
AWS IoT Integration & TLS Support in MQTT

diwanc:
**TLS Feature**
- Adding TLS support and certificate validations in current MQTT implementation
- Adding identity layer to encapsulate gateway identity and TLS configurations
- Making keepalive parameter configurable
- Updating client certification validation method
- Adding sample MQTT configurations

**AWS IoT DCC**
- DCC for AWS leveraging [AWS IoT SDK](https://github.com/aws/aws-iot-device-sdk-python)
- MessagingAttributes class for holding transport specific messaging parameters Eg: Topic, QoS etc.,
- Example to publish data from LIOTA to AWS IOT platform via AWS DCC


example/aws_iot/home_temp_simulated.py
```diff
@@ -0,0 +1,108 @@
...
+
+if __name__ == '__main__':
+    gateway = Dk300(config['Gateway1Name'])
+
+    #  Using Identity layer for holding details related to authentication
+    identity = Identity(config['RootCAPath'], config['ClientCertPath'], config['PrivateKeyPath'], None, None)
+
+    #  Connecting to AWS
+    #  These are standard steps to follow to connect with AWS IoT using AWS IoT SDK
+    mqtt_client = AWSIoTMQTTClient(config['ClientName'])
```
pmasrani: I guess by defining a common DCC layer, we are trying to avoid these exact steps. All these operations can be performed within AWSIoT.__init__() method so that the end client code is simplified.

example/sampleProp.conf
```diff
@@ -7,6 +7,33 @@ GraphiteIP = "Graphite-IP"
 GraphitePort = None
 Device1Name = "Device-Name"
 
+#MQTT Configurations
```
pmasrani: This file is growing too big and most of the parameters can be irrelevant to other examples. We can create sections like [AWS], [VROPS] etc or can make altogether new config file say, aws_sampleProp.conf if nothing is shared.

diwanc: Will move AWS props to a new conf file

example/aws_iot/home_temp_simulated.py
diwanc: Makes sense. Will look at optimizing this.

diwanc:
Will leverage refatored codebase for further commits. Closing this.

PR #21, opened 2016-08-23
Gateway name made part of the uuid.ini file

KohliDev:
Gateway name with uuid made part of the uuid.ini file


bfrggit:
Looks good.

liota/dcc/vrops.py
```diff
@@ -128,8 +128,10 @@ def on_receive_safe(msg):
                             try:
                                 uuid_path = config.get('UUID_PATH', 'uuid_path')
```
pmasrani: Need to rename UUID_PATH and uuid.ini filename as it contains more details now, not just UUID.

bfrggit: Well... Just same details split into two lines...

PR #22, opened 2016-08-23
modifying metric values storage from list to queue

winniex1:
changed for both vrops and graphite;
also fixed current_aggregation_size reset issue: 
now in collection thread, when send_queue.put(matric), reset that conter.

### DIFF
--- a/liota/dcc/vrops.py
+++ b/liota/dcc/vrops.py
@@ -151,21 +151,15 @@ class Vrops(DataCenterComponent):
     def subscribe(self):
         pass
 
-    def publish(self, metric):
-        met_cnt = metric.values.qsize()
-        if met_cnt == 0:
-            return
-        timestamps = []
-        values = []
-        for _ in range (met_cnt):
-            m = metric.values.get(block = True)
-            if m is not None:
-                timestamps.append(m[0])
-                values.append(m[1])
-        if timestamps == []:
-            return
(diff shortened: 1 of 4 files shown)


liota/dcc/vrops.py
```diff
@@ -150,9 +150,20 @@ def subscribe(self):
         pass
 
     def publish(self, metric):
-        timestamps = [t for t, _ in metric.values]
-        values = [v for _, v in metric.values]
+        met_cnt = metric.values.qsize()
+        if met_cnt == 0:
+            return
```
bfrggit: Since we already talked about this, I am gonna be brief.
`dcc.publish` should take a list, e.g. `dcc.publish(self, metric_values)`, the list will be a list of tuples got from the queue (`metric.values`).
`metric.send_data` is gonna retrieve data from the queue (`self.values`). That means before it calls `self.dcc.publish` it should count the number of items in the queue, and do so many number of `self.values.get( )` to get exactly the same amount of numbers, and put these things into a list, and finally call `self.dcc.publish(that_list)`.
Then, dcc codes can stay ALMOST unchanged compared to the original implementation.

bfrggit:
Will not change Winnie's codes for the publish() interface for now, because changing that ultimately requires refactoring LIOTA's gateway/DCC codes.

liota/dcc/vrops.py
bfrggit: Will not make this change for now.

PR #24, opened 2016-09-18
End-to-end simulated-system to Graphite flow in Refactor branch

pmasrani:


liota/entities/metrics/registered_metric.py
```diff
@@ -30,39 +30,102 @@
...
 
     def __init__(self, ref_metric, ref_dcc, reg_entity_id):
         if not isinstance(ref_metric, Metric):
             raise TypeError
-        super(self.__class__, self).__init__(
-            ref_entity=ref_metric,
-            ref_dcc=ref_dcc,
-            reg_entity_id=reg_entity_id
-        )
+        RegisteredEntity.__init__(self,
```
bfrggit: What is the advantage of using `ParentClass.__init__(self)` against `super(self.__class__, self).__init__()`?

liota/core/metric_handler.py
```diff
@@ -30,4 +30,177 @@
...
+        Thread.__init__(self)
+        self.daemon = True
+        self.start()
+
+    def run(self):
+        global event_ds
+        global collect_queue
+        global send_queue
+        while True:
+            metric = collect_queue.get()
```
bfrggit: Must check if the metric is marked as "deleted" (there should be a `flag_alive`).

liota/core/metric_handler.py
```diff
@@ -30,4 +30,177 @@
...
 
-# TODO: Implement this module
+from Queue import Queue, PriorityQueue, Full
+import heapq
+import logging
+from threading import Thread, Condition
+from time import time as _time
+
+from liota.lib.utilities.utility import getUTCmillis
+
```
bfrggit: This file lacks necessary updates in my package manager fork. What I had there may not be the best or even a good implementation but better have some sort of implementation or package manager may not work correctly later.

And also, think of a way to close these threads elegantly.

liota/entities/metrics/registered_metric.py
```diff
@@ -30,39 +30,102 @@
...
+        metric_handler.event_ds.put_and_notify(self)
 
     def stop_collecting(self):
         self.flag_alive = False
+        raise NotImplementedError
+
+    def write_full(self, t, v):
+        self.values.put((t, v))
+
+    def write_map_values(self, v):
```
bfrggit: What is the "map"???

liota/entities/metrics/registered_metric.py
```diff
@@ -30,39 +30,102 @@
...
+        log.debug("self.aggregation_size:" +
+                  str(self.ref_entity.aggregation_size))
+        return self.current_aggregation_size >= self.ref_entity.aggregation_size
+
+    def collect(self):
+        log.debug("Collecting values for the resource {0} ".format(
+            self.ref_entity.name))
+        self.args_required = len(inspect.getargspec(
+            self.ref_entity.sampling_function)[0])
+        if self.args_required is not 0:
```
bfrggit: This part was a pretty confusing part in the previous code, too! Why do we pass `1` as the argument if `args_required is not 0`?

liota/lib/utilities/utility.py
```diff
@@ -0,0 +1,146 @@
...
+
+from datetime import datetime
+import hashlib
+import logging
+import os
+import platform
+import random
+import uuid
+
+
```
bfrggit: `utility.py` is very bad naming. Everything here is some sort of utility. Find some good names and separate irrelevant functions.

PR #26, opened 2016-09-21
add Dell5k/graphite related implementations; port package manager; add bike/thermistor as simulated devices (also add examples)

winniex1:
1. add Dell5K implementation, and create an example based on it
2. port Package Manager functionality (modify metric_handlers, add package_manager)
3. add Bike model and Thermistor models as devices, and create examples based on them
4. add packages examples based on system, bike and thermistor devices
5. add above into setup.py
6. modify system/devices/metrics to use super() based on Vaibhav’s comments


examples/de5k_system_graphite.py
```diff
@@ -0,0 +1,72 @@
...
+config.readfp(open('sampleProp.conf'))
+
+# Random number generator, simulating random metric readings.
+
+
+def simulated_sampling_function():
+    return random.randint(0, 20)
+
+# ---------------------------------------------------------------------------
+# In this example, we demonstrate how data for a simulated metric generating
```
pmasrani: Please change the comments to indicate that this program should be run on Dell Edge Gateway 5000. Can we add some functionality for that? Simulated example is more for a quick start of Liota on Ubuntu/Windows. Can we read some metrics from Dell Edge Gateway?

examples/model_simulated/graphite_bike_simulated.py
```diff
@@ -0,0 +1,178 @@
...
+#  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR        #
+#  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF       #
+#  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS   #
+#  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN    #
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)    #
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF     #
+#  THE POSSIBILITY OF SUCH DAMAGE.                                            #
+# ----------------------------------------------------------------------------#
+
+import ConfigParser, time, math, pint
```
pmasrani: Imports in incorrect order and format. If you are using Pydev in Eclipse, it can easily be corrected by Ctrl+Shift+O.

examples/model_simulated/vrops_bike_model_simulated.py
```diff
@@ -0,0 +1,188 @@
+# -*- coding: utf-8 -*-
```
pmasrani: Let's use this refactoring opportunity to remove all references to vRops. Rename it to Vmware_Ice

examples/model_simulated/vrops_thermistor_model_simulated.py
```diff
@@ -0,0 +1,158 @@
...
+if __name__ == '__main__':
+
+    # create a data center object, vROps in this case, using websocket as a transport layer
+    # this object encapsulates the formats and protocols neccessary for the agent to interact with the dcc
+    # UID/PASS login for now.
+    vrops = Vrops(config['vROpsUID'], config['vROpsPass'], WebSocket(url=config['WebSocketUrl']))
+
+    # create a gateway object encapsulating the particulars of a gateway/board
+    # argument is the name of this gateway
+    gateway = Dk300(config['Gateway1Name'])
```
pmasrani: Are you able to successfully run this program? We have changed sampleProp.conf to have a [SECTION] and are reading it using ConfigParser in other sample programs.

examples/sampleProp.conf
```diff
@@ -3,6 +3,7 @@ GatewayName = Gateway-Name
 MetricName = Metric-Name
 MetricWithOwnTsName = Metric-With-Own-Ts-Name
 BulkCollectedMetricName = Bulk-Collected-Metric-Name
+DeviceName = Gateway-Name
```
pmasrani: Can you move this up the order. Generally one creates a Gateway, a Device and then a Metric. Also, change its default value to Device-Name.

liota/core/package_manager.py
```diff
@@ -30,4 +30,1104 @@
...
+    try:
+        while not isinstance(
+                actual_package_manager.package_thread,
+                actual_package_manager.PackageThread
+            ) or actual_package_manager.package_thread.isAlive():
+            sleep(1)
+    except (KeyboardInterrupt, SystemExit):
+        pass
+    finally:
+        if not actual_package_manager.package_message_queue is None:
```
pmasrani: (if <condition> is not None) will be much more readable.
http://stackoverflow.com/questions/2710940/python-if-x-is-not-none-or-if-not-x-is-none

liota/core/package_manager.py
```diff
@@ -30,4 +30,1104 @@
...
+            actual_package_manager.package_message_queue.put(["terminate"])
+    log.info("MainThread exits")
```
pmasrani: "Exiting Package Manager"?

packages/sampleProp.conf
```diff
@@ -0,0 +1,23 @@
...
+Metric3Name = Metric3-Name
+Metric4Name = Metric4-Name
+
+GatewayPropList = {"Country":"USA-G", "State":"California", "City":"Palo Alto", "Location":"VMware HQ", "Building":"Promontory H Lab", "Floor":"First Floor"}
+
+[GRAPHITE]
+IP = Graphite-IP-Address
+Port = Graphite-Port
+
+[BIKE_SIMULATOR]
```
pmasrani: My idea of implementing sections was to be easily able to club properties related to DCC, DCCComms, DeviceComms, DefaultPropertiesForAllPrograms etc. I don't think we should have different sections for each simulator. Why can't the Bike Simulator or Thermistor Simulator reuse the properties from the DEFAULT section?

liota/core/package_manager.py
bfrggit: This part is likely to be a fallback option when someone runs PackageManager as the MainThread using `python -m whatever.ref.to.package_manager`. It is expected to operate in the same way as if he is running it with `liotad.py`. Therefore, it is actually MainThread exiting.

bfrggit: Python won't import the same module twice **when it is imported using the same name**. When running `python -m`, the module being added for `-m` option is loaded as `__main__`.

I know normally people won't run it this way. By writing this part of code, I was trying to avoid a corner case where it can cause package manager to be loaded twice (once as `__main__` and once as `liota.core.package_manager`).

Better keep it as "MainThread exits" to avoid confusion with "PackageThread" and "PackageMessengerThread".

liota/core/package_manager.py
```diff
@@ -30,4 +30,1104 @@
...
+        self._packages_loaded[file_name] = package_record
+
+        log.info("Package class from module %s is initialized" \
+                % module_loaded.__name__)
+        return package_record
+
+    #-----------------------------------------------------------------------
+    # This method is called to unload package using its file_name (no ext).
+
+    def _package_unload(self, file_name, track_list=None):
```
pmasrani: Need to define the parameters like track_list. I didn't understand its use until i reached package_reload.

liota/core/package_manager.py
```diff
@@ -30,4 +30,1104 @@
...
+        self._resource_registry.register("package_conf", package_path)
+        self.flag_alive = True
+        self.start()
+
+    #-----------------------------------------------------------------------
+    # This method is used to handle listing commands
+
+    def _cmd_handler_list(self, parameter):
+        if parameter == "packages" or parameter == "pkg":
+            log.warning("List of packages - \n\t%s" \
```
winniex1: we do want to these printed out in liota.log since this is result of these cmds (e.g., list pkg...). 
Our default message level is warning, so  better to be warning.

liota/core/package_manager.py
```diff
@@ -30,4 +30,1104 @@
...
+    # file_name (no_ext) as package identifier.
+
+    def _package_load(self, file_name, ext_forced=None, check_stack=None):
+        global package_path
+
+        log.debug("Attempting to load package: %s" % file_name)
+
+        # Check if specified package is already loaded
+        if file_name in self._packages_loaded:
+            log.warning("Package already loaded: %s" % file_name)
```
winniex1: not really error. so better keep as warning.

liota/core/package_manager.py
```diff
@@ -30,4 +30,1104 @@
...
+        package_record = self._packages_loaded[file_name]
+        assert(isinstance(package_record, PackageRecord))
+
+        # Stop all dependents, before making any change to current package
+        dependents = package_record.get_dependents()
+        
+        if len(dependents) > 0:
+            log.info("Package %s is depended by: %s" \
+                    % (file_name, " ".join(dependents)))
+            for dependent in dependents:
```
winniex1: just need to mention package 2 (when load package 2, package manager will check package 1).
From package examples, it's easy to realize it.

liota/core/package_manager.py
pmasrani: We should not be changing the log levels for displaying it in liota.log. Instead, the default logging level for liota needs to be changed. Currently, even when i start a Liota program, there is no way the user comes to know about it. I talked to Vaibhav about it yesterday, and we think it will be good to change the default logging level to INFO. We will definitely have to change the current INFO logs to DEBUG as most of them are useful only while debugging and keep only the most useful ones at INFO level.

PR #27, opened 2016-09-21
Submit modifications for last commit based on review comments

winniex1:


examples/dell5k_system_graphite.py
```diff
@@ -31,42 +31,45 @@
 # ----------------------------------------------------------------------------#
 
 import ConfigParser
-import random
-
+from linux_metrics import cpu_stat
```
pmasrani: Let us use psutil instead of linux_metrics as it is a cross-platform library.\ and we want to support windows going ahead.

PR #28, opened 2016-09-21
More fixes based on review commens; use cross-platform psutil to get system info

winniex1:


bfrggit:
In `package_manager.py`, at the beginning of the file, like Ln 50-70 something, there was a "MainThread" being changed to "Package manager". This was not very clear actually.

I said "MainThread exiting", etc. because that part of code was used to handle the case where sb. runs `python -m liota.core.package_manager` so Package Manager runs as the "MainThread". In this case, MainThread doesn't really init a package manager, but import `package_manager`, like what `liotad.py` does. Therefore, that's actually just MainThread exiting, not the really functional package manager...

winniex1:
Ok. Will change back.

PR #31, opened 2016-09-25
Addition of Gateway, Devices, Websocket communication & IOTCC DCC

KohliDev:
Still working on the examples folder will add all the examples
together in the next review request.Improved code at multiple places will affect the current checked in examples will take care of them as well.


liota/__init__.py
```diff
@@ -94,21 +94,21 @@ def setup_package():
...
-                pass
-            else:
-                raise
+        # if not os.path.exists(dir):
+        #     log.warn('directory does not exist; create it')
+        #     try:
+        #         supermakedirs(dir, 777)
+        #     except Exception as e:
+        #         raise OSError('Cannot create directory' + dir)
+        #     finally:
```
KohliDev: This piece of code didn't work on my machine. Is it tested ?  As it gave reference error.
And what is the usage of giving permission as 777 (we are making the directory in user machine permission should be 644)

winniex1:
Yes, definitely have tested it. I record original permission, and after creating directory will reset it back.
777 is used since want to make directory with super mode to guarantee success.

Before generate fifo, there is chance that no directory there. We need to create directory.
Did you use Ubuntu to test?

bfrggit:
If the purpose of that piece of code is to generate the FIFO file -- if I understood it correctly, then here is my code in the old package manager implementation that does the same thing: https://github.com/bfrggit/liota/blob/master/liota/core/package_manager.py, starting Ln 1068, ending Ln 1095.

Permissions are better to be carefully designed, instead of being set to 777 to workaround possible failure. I don't think it is necessary for a directory to have "w" permission for "others".

Also, why do we move these checks from package manager to `liota/__init__.py`? If somebody decides to run liota without package manager (as was suggested to be supported, by @pmasrani), then there's no reason we create these paths and pipe.

liota/dccs/iotcc.py
```diff
@@ -0,0 +1,248 @@
...
+                else:
+                    raise IOError('Could not open config file ' + fullPath)
+            except IOError, err:
+                log.error('Could not open config file')
+        else:
+            # missing config file
+            log.warn('liota.conf file missing')
+
+    def publish(self, reg_metric):
+        met_cnt = reg_metric.values.qsize()
```
pmasrani: Publish functionality in dcc.py made sense. Before publishing a metric, it should be formatted and then send on the con object should be called. Instead of performing all these conversions in the publish() method, do it inside _format_data() method.

liota/__init__.py
pmasrani: I agree with Vaibhav. It is never a good practice to use permissions of 777.

liota/dccs/dcc.py
```diff
@@ -69,9 +68,9 @@ def _create_relationship(self, entity_parent, entity_child):
     def _format_data(self, reg_metric):
         pass
 
+    @abstractmethod
     def publish(self, reg_metric):
-        message = self._format_data(reg_metric)
```
pmasrani: Keep this code. Publish should have two functionalities: Formatting and sending.

liota/dccs/iotcc.py
```diff
@@ -0,0 +1,248 @@
+# -*- coding: utf-8 -*-
```
pmasrani: We are using the name Project ICE for all promotions in vmworld and other road shows. Why name it iotcc now?

liota/dccs/iotcc.py
```diff
@@ -0,0 +1,248 @@
...
+from liota.entities.metrics.metric import Metric
+from liota.utilities.utility import getUTCmillis, LiotaConfigPath
+from liota.utilities.si_unit import parse_unit
+from liota.entities.registered_entity import RegisteredEntity
+
+log = logging.getLogger(__name__)
+
+
+class IotControlCenter(DataCenterComponent):
+    """ The implementation of vROPS cloud provider solution
```
pmasrani: vrops?

liota/dccs/iotcc.py
```diff
@@ -0,0 +1,248 @@
...
+                        self.proto.on_receive(json.loads(msg))
+                        log.debug("Processed msg: {0}".format(json_msg["type"]))
+                        if json_msg["type"] == "create_or_find_resource_response":
+                            if json_msg["body"]["uuid"] != "null":
+                                log.info("FOUND RESOURCE: {0}".format(json_msg["body"]["uuid"]))
+                                self.reg_entity_id = json_msg["body"]["uuid"]
+                                exit()
+                            else:
+                                log.info("Waiting for resource creation")
+                                time.sleep(5)
```
pmasrani: Do we need this sleep call?

liota/dccs/iotcc.py
```diff
@@ -0,0 +1,248 @@
...
+        config = ConfigParser.RawConfigParser()
+        fullPath = LiotaConfigPath().get_liota_fullpath()
+        if fullPath != '':
+            try:
+                if config.read(fullPath) != []:
+                    try:
+                        uuid_path = config.get('UUID_PATH', 'uuid_path')
+                        uuid_config = ConfigParser.RawConfigParser()
+                        uuid_config.optionxform = str
+                        uuid_config.add_section('GATEWAY')
```
pmasrani: No GATEWAY anymore. SYSTEM?

liota/dccs/iotcc.py
```diff
@@ -0,0 +1,248 @@
...
+                reg_metric.timestamps.append(m[0])
+                reg_metric._values.append(m[1])
+        if reg_metric.timestamps == []:
+            return
+        message = self._format_data(reg_metric)
+        log.info('publishing {0}'.format(message))
+        self.con.send(message)
+
+    def _create_relationship(self, entity_parent_reg_id, entity_child_reg_id):
+        """ This function initializes all relations between gateway and it's children.
```
pmasrani: No gateway

liota/entities/devices/RAM.py
```diff
@@ -30,4 +30,31 @@
...
+from abc import ABCMeta, abstractmethod
+from liota.entities.devices.device import Device
+from liota.dccs.dcc import DataCenterComponent
+from liota.utilities.utility import systemUUID
+
+
+class Ram(Device):
+
+    """
+    Abstract base class for all devices (things).
```
pmasrani: Incorrect comment?

liota/entities/systems/dk300_system.py
```diff
@@ -0,0 +1,8 @@
+from system import System
+from liota.utilities.utility import systemUUID
+
+
+class Dk300System(System):
+
+    def __init__(self, name):
+        super(Dk300System, self).__init__(name=name, entity_id=systemUUID().get_uuid(name))
```
pmasrani: If entity_id is always generated, let us move this call to the base Entity call. Let us not call this from every class we are defining.

setup.py
```diff
@@ -45,7 +45,7 @@
 
 setup(
     name='liota',
-    version='0.1.3',
+    version='0.1.4',
```
pmasrani: The version in master branch is 0.1.2. I changed it to 0.1.3 for the refactoring branch. Let's keep it to 0.1.3 for the whole refactoring code.

liota/entities/devices/RAM.py
```diff
@@ -30,4 +30,31 @@
 #  THE POSSIBILITY OF SUCH DAMAGE.                                            #
 # ----------------------------------------------------------------------------#
 
-# TODO: Implement this module
+from abc import ABCMeta, abstractmethod
```
pmasrani: Is this example providing any value add? I see that it is just extending the device class as-is. In the current form, it is not even changing the entity_type. I think we should add it only to demonstrate additional functionality.

liota/entities/devices/device.py
```diff
@@ -59,11 +59,11 @@ def __init__(self, name, parent, entity_id, entity_type="Device"):
             entity_type=entity_type
         )
 
-    def register(self, entity_obj, dcc_obj, reg_entity_id):
+    def register(self, dcc_obj, reg_entity_id):
```
pmasrani: Ideally I should have commented this in Winnie's review. Somehow, the register() call in the base class like device seems to be misplaced. We separated out the entities and registered_entities. Now we are having the register() call in entity?

liota/lib/transports/web_socket.py
```diff
@@ -57,84 +57,81 @@ def connect_soc(self):
             log.exception("WebSocket exception, please check the WebSocket address and try again.")
             sys.exit(0)
 
-
-    def WebSocketConnection(self, host, verify_cert=True, CERTDIR="/var/lib/helix-agent/cert"):
-
-      if not verify_cert:
+    # CERTPATH to be taken in consideration later
```
pmasrani: Add it as a TODO

liota/__init__.py
```diff
@@ -94,21 +94,21 @@ def setup_package():
...
-            finally:
-                os.umask(original_umask)
-        try:
-            os.mkfifo(path)
-        except OSError as exc:  # Python > 2.5
-            if exc.errno == errno.EEXIST:
-                pass
-            else:
-                raise
+        # if not os.path.exists(dir):
```
pmasrani: remove the commented code once we finalize this implementation.

liota/dccs/iotcc.py
```diff
@@ -0,0 +1,248 @@
...
+            self.con.on_receive = on_receive_safe
+            thread.daemon = True
+            thread.start()
+            self.con.send(
+                self.registration(self.con.next_id(), entity_obj.entity_id, entity_obj.name, entity_obj.entity_type))
+            thread.join()
+            log.info("Resource Registered {0}".format(entity_obj.name))
+            if entity_obj.parent is not None:
+                self._create_relationship(entity_obj.parent.reg_entity_id, self.reg_entity_id)
+                log.info("Relationship Created")
```
pmasrani: Need better logs, relationship created between which resources?

liota/dccs/iotcc.py
```diff
@@ -0,0 +1,248 @@
...
+    def _create_relationship(self, entity_parent_reg_id, entity_child_reg_id):
+        """ This function initializes all relations between gateway and it's children.
+            It is called after each object's UUID is received.
+
+            Parameters:
+            - obj: The object that has just obtained an UUID
+        """
+        self.con.send(self.relationship(self.con.next_id(), entity_parent_reg_id, entity_child_reg_id))
+
+    def registration(self, msg_id, res_id, res_name, res_kind):
```
pmasrani: Most of these functions need to start with _

liota/lib/transports/web_socket.py
```diff
@@ -57,84 +57,81 @@ def connect_soc(self):
...
-              self.close()
-              os._exit(0)
+        request_calls = ['request', 'response']
+        complete_message = json.dumps(msg)
+        log.info("TX Sending message {0}".format(complete_message))
+        try:
+            self.ws.send(complete_message)
+        except:
+            # Retry logic only for publishing stats, not for request or response calls
+            if all(request not in complete_message for request in request_calls):
```
pmasrani: Can we extract this retry login and make it generic for transport layers?

liota/dccs/dcc.py
KohliDev: Publish should be an abstract method as it has to be overidden by the base class and implementation goes in the base class.

liota/dccs/iotcc.py
KohliDev: Yes, the naming mechanism has to be decided together.IoTControlEnterprise makes sense then ProjectIce in the liota code base.

liota/dccs/iotcc.py
KohliDev: It is a wait call to check back if the resource is created or not. I can remove this time.sleep but then the too many request will fire up from the liota to IoTCC.

liota/dccs/iotcc.py
KohliDev: JIRA task created already.

liota/entities/devices/device.py
KohliDev: Register is an abstract method in the base class will have to be overridden in the child class and better to keep the implementation in device.py as the register implementation might differ from system.

liota/lib/transports/web_socket.py
KohliDev: Can be done but over here we have specific self.ws.send used to publish the message after the websocket object is re-created (can be revisited later).

liota/__init__.py
KohliDev: As per @bfrggit comment, this code  should be moved from __init__.py to package manager so that we can have the functionality of package manager separate from the normal code. Also the permission should be revisited.

liota/entities/systems/dk300_system.py
KohliDev: Something has to be passed from the child class as it is a parameter. So better to generate the UUID here.

PR #34, opened 2016-09-27
Additional code changes to include example and refactor the code

KohliDev:
1.  Refactored the code to keep the map of the entity_id & reg_entity_id in the uuid.ini file.
    
2. Removed entity_obj.register call from the Graphite & IoTCC DCC because directly RegisteredMetric and RegisteredEntity Objects can be created.

3. Included the new samplePropIce config file in order to read the properties list using config exec file.
    
4. Removed publish method from Graphite DCC as it will be kept only in the base dcc class.


examples/dell5k_system_graphite.py
```diff
@@ -35,7 +35,7 @@
 from liota.dcc_comms.socket_comms import Socket
 from liota.dccs.graphite import Graphite
 from liota.entities.metrics.metric import Metric
-from liota.entities.systems.de5k_system import Dell5KSystem
+from liota.entities.systems.dell5k_system import Dell5KSystem
```
KohliDev: The example failed in my testing, I figured out that import statement was incorrect, fixed in this review request.

PR #36, opened 2016-09-29
 More modifications

winniex1:
1. use dcc's create_relationship() to setup parent<->children;
relationships are only between RegisteredEntity or RegisteredMetric;
so Entity does not need to hold parent reference;
RegisteredEntity will keep its uuid in dcc;
RegisteredMetric will keep its parent's uuid in dcc
2. modify dcc's register() to decouple creating resource in dcc and
create relationship for registered entities
3. to work with AW agent, store EdgeSystem and Devices names and uuid
in iotcc.json file
4. use EdgeSystem everywhere
5. To keep EdgeSystems under Helix IoT Adaptor in vRops, use "HelixGateay"
temporarily as EdgeSystem's resource kind. Later Adaptor side needs
accept "EdgeSystem" resource kind also (also put it under Adaptor).
6. modify all related examples to work with above modifications
7. modify all package manager related examples to work


examples/multi_metric_simulated_edgesystem_graphite.py
```diff
@@ -72,32 +72,32 @@ def simulated_list_of_timestamps_values_sampling_function():
 
 if __name__ == '__main__':
 
-    system = SimulatedSystem(config['SystemName'])
+    edgesystem = SimulatedEdgeSystem(config['EdgeSystemName'])
```
KohliDev: variable name can be changed to edge_system up to you.

examples/simulated_edgesystem_graphite.py
```diff
@@ -55,16 +55,17 @@ def simulated_sampling_function():
 
 if __name__ == '__main__':
 
-    system = SimulatedSystem(config['SystemName'])
+    edgesystem = SimulatedEdgeSystem(config['EdgeSystemName'])
```
KohliDev: Variable name can be changed to edge_system everywhere better readability of code.

liota/entities/registered_entity.py
```diff
@@ -36,6 +36,16 @@ def __init__(self, ref_entity, ref_dcc, reg_entity_id):
         self.ref_entity = ref_entity
         self.ref_dcc = ref_dcc
         self.reg_entity_id = reg_entity_id
+        self.parent = None
```
KohliDev: I don't see the requirement of defining parent attribute and the set_parent method. We can just directly create the relationship as per yesterday's discussion we don't require the parent to be part of the registered_entity. We can directly make the relationship call with help of reg_entity_id and reg_child_id. In the case of metric as of know, you'll have to set the reg_entity_id of the parent to reg_metric_entity_id so that we can invoke start_collecting method on it.

liota/dccs/iotcc.py
```diff
@@ -125,79 +124,34 @@ def on_receive_safe(msg):
...
+        if (not isinstance(reg_entity_parent, RegisteredEntity) \
+            and not isinstance(reg_entity_parent, RegisteredMetric)) \
+            or (not isinstance(reg_entity_child, RegisteredEntity) \
+                and not isinstance(reg_entity_child, RegisteredMetric)):
+            raise TypeError()
+
+        reg_entity_child.set_parent(reg_entity_parent)
+        if isinstance(reg_entity_child, RegisteredMetric):
+            # should save parent's reg_entity_id
+            reg_entity_child.set_reg_entity_id(reg_entity_parent.get_reg_entity_id())
```
KohliDev: You don`t require getter & setter in python. Just directly call the object attribute.

packages/examples/iotcc_edgesystem_stats.py
```diff
@@ -78,43 +78,47 @@ def run(self, registry):
 
         # Create metrics
         self.metrics = []
-        metric_name = "system.CPU_Utilization"
-        metric1 = Metric(name=metric_name, parent=system,
+        metric_name = "edgesystem.CPU_Utilization"
+        metric1 = Metric(name=metric_name,
```
KohliDev: Please rename the variable metric1 correctly metric_cpu_utilization.

liota/entities/registered_entity.py
winniex1: Consider in the future, Metric could be the parent of another parent. parent will be a common thing for both RegisteredEntity and RegisteredMetric. It seems to me correct to put parent in the base class of RegistereEntity. 

I agree that in python we can directly in other module call object attribute even in base class. I'll do that way.

winniex1: No, I do need to locally hold the relationship. Previously, we do not define it in RegisteredEntity since Entity has it. We can trace there.

When set properties, in RegisteredMetric case, reg_entity_obj.parent is used. 
Can you remember that previously AWS DCC checkin's want to let Metrics have attributes?
We can make it as properties. So it is needed.

winniex1: not chekin (I mean their pull request)

KohliDev: Okay, this is reasonable. 

entity.entity_type, entity.name (in logs) is the only place where the parent object is required in the set_properties method. We can actually pass entity_type as an attribute like we set the reg_entity_id during create_relationship call. In future, when we register Metric as a separate Entity we won`t require this code as property would be set in the similar way they are set on the device or gateway. 

Also, AWS DCC if they want to set attribute(property) they will set them directly on the metric object rather then on the parent.

liota/dccs/iotcc.py
bfrggit: You might be right... Not very sure which is better practice. Setters might provide flexibility in possible future implementation (maybe not for this specific instance).

examples/simulated_edgesystem_graphite.py
bfrggit: If **EdgeSystem** is the camel case, then the corresponding variable name should be **edge_system**, saying they both contain two words.

PR #37, opened 2016-09-30
Check in followings:

winniex1:
1. Let EdgeSystem entity_type be "EdgeSystem"; iotcc will make internal conversion
to put EdgeSystem under IoT Adaptor
2. Use edge_system rather than edgesystem
3. Write EdgeSystem name and uuid to uuid.ini file to keep AW agent unchanged
4. Add conditional check for each register call in examples
5. IotCC directly get/set RegisteredEngity/RegisteredMetric attributes
rather than using get/set call in RegisteredEntity/RegisteredMetric
6. Make a create_relationship() for Graphite also, where child only record parent locally and no
DCC operations, and used in examples
7. Delete _ in Metric Name in IoTCC examples, since vRops 6.3.1 cannot show them
8. other modifications based on review comments


examples/dk300_edge_system_iotcc.py
```diff
@@ -172,27 +188,36 @@ def simulated_device():
     #        another Resource in IoTCC of which the should be the child of a parent-child relationship among Resources
     ram_device = SimulatedDevice(config['DeviceName'], "Device-RAM")
     reg_ram_device = iotcc.register(ram_device)
-    iotcc.create_relationship(reg_edgesystem, reg_ram_device)
+    if reg_ram_device is None:
+        print "Device registration to IOTCC failed"
+        exit()
```
GregBollella: I don't get this. Why if we could not register a device do we have the program exit?

examples/model_simulated/graphite_bike_simulated.py
```diff
@@ -161,6 +161,9 @@ def get_bike_power():
     graphite = Graphite(Socket(ip=config['GraphiteIP'],
                                port=config['GraphitePort']))
     graphite_reg_dev = graphite.register(bike_model)
+    if graphite_reg_dev is None:
+        print "Device registration to Graphite failed"
+        exit()
```
GregBollella: I don't like exit()'s here. Especially without comments.

examples/model_simulated/graphite_thermistor_simulated.py
```diff
@@ -130,6 +130,9 @@ def get_thermistor_temperature():
                                port=config['GraphitePort']))
 
     graphite_reg_dev = graphite.register(thermistor_model)
+    if graphite_reg_dev is None:
+        print "Device registration to Graphite failed"
+        exit()
```
GregBollella: I prefer something like 
if graphite_reg_dev:
       do all the stuff here
else:
      report error

winniex1:
Yes, can modify it.

PR #47, opened 2016-10-12
AWSIoT DCC, DCCComms and Example

Venkat2811:
`AWSMQTTDccComms:`
- Uses [AWSIoTPythonSDK](https://github.com/aws/aws-iot-device-sdk-python).
- `init()` accepts AWSIoT object because, there are lot of parameters to be handled to enclose it as part of init() apart from credentials. Refer [here](https://github.com/aws/aws-iot-device-sdk-python/blob/master/samples/basicPubSub/basicPubSub.py#L129).  By using it this way,  DCC would be compatible with future versions AWS SDK also.

`AWSIoT DCC:` 
- Publish topic is generated based on Parent-Child hierarchy.  
- Added register_entity() to call create_relationship() as part of registration itself to make it much more easier for clients.
- Message published in JSON format.

`AWSIoT Package Manager:`
- Added.

`Unit & Integration TestSuites:`
- Will be added once this PR is accepted.

### DIFF
--- a/examples/aws_iot/home_simulated.py
+++ b/examples/aws_iot/home_simulated.py
@@ -33,7 +33,7 @@
 from liota.dcc_comms.aws_mqtt_dcc_comms import AWSMQTTDccComms
 from liota.dccs.aws_iot import AWSIoT
 from liota.entities.metrics.metric import Metric
-from liota.entities.devices.device import Device
+from liota.entities.devices.simulated_device import SimulatedDevice
 from AWSIoTPythonSDK.MQTTLib import AWSIoTMQTTClient
 from liota.entities.edge_systems.dell5k_edge_system import Dell5KEdgeSystem
 import random
@@ -134,8 +134,8 @@ if __name__ == '__main__':
     #  at topic 'TestGatewayName/CPUUtilization'
     reg_cpu_utilization.start_collecting()
 
-    #  Creating Device
-    dht_sensor = Device("SimulatedDHTSensor", None)
+    #  Creating Simulated Device
+    dht_sensor = SimulatedDevice("SimulatedDHTSensor")
     #  Registering Device and creating Parent-Child relationship
     reg_dht_sensor = aws.register_entity(reg_edge_system, dht_sensor)
     #  Creating Temperature Metric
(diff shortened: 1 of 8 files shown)


examples/aws_iot/home_simulated.py
```diff
@@ -0,0 +1,190 @@
...
+    return round(psutil.cpu_percent(interval=sample_duration_sec), 2)
+
+
+#  Random number generator, simulating living room temperature readings.
+def living_room_temperature():
+    return random.randint(10, 30)
+
+
+#  Random number generator, simulating living room humidity readings.
+def living_room_humidity():
```
KohliDev: Very Good Example

liota/dcc_comms/aws_mqtt_dcc_comms.py
```diff
@@ -0,0 +1,99 @@
...
+#  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR        #
+#  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF       #
+#  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS   #
+#  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN    #
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)    #
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF     #
+#  THE POSSIBILITY OF SUCH DAMAGE.                                            #
+# ----------------------------------------------------------------------------#
+
+import logging
```
KohliDev: Can we leverage the normal MQTT implemented in the previous PR and use it in spite of creating a separate aws_mqtt_dcc_comms.

I can understand that in order to show the usecase around "AWSIoTMQTTShadowClient" this code must have been added. But, I`ll like to keep the below simple implementation in liota if possible:

MQTT_general ---> AWS

And if the AWS shadow capabilities are required to be demonstrated then lets add it in the user program as it will be specific to AWS example only.

examples/aws_iot/home_simulated.py
Venkat2811: Thanks :)

liota/dcc_comms/aws_mqtt_dcc_comms.py
Venkat2811: We'll discuss about this.

Venkat2811:
Closing this as we decided not to use SDK

PR #49, opened 2016-11-07
RangeFilter

Venkat2811:
- Abstract Filter class
- WindowingScheme class.
- RangeFilter with and without Windowing scheme


liota/lib/utilities/filters/windowed_filters/windowed_range_filter.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+
+    def __init__(self, filter_type=None, lower_bound=None, upper_bound=None, window_size_sec=10):
+        """
+        :param filter_type: Any one type from Filter Enum of RangeFilter.
+        :param lower_bound: Lower bound
+        :param upper_bound: Upper bound
+        :param window_size_sec: Configurable time window for heartbeat
+        """
+        RangeFilter.__init__(self, filter_type, lower_bound, upper_bound)
```
KohliDev: Use Super

examples/dell5k_edge_system_with_filter_iotcc.py
```diff
@@ -0,0 +1,203 @@
...
+    return net_usage_filter.filter(round(psutil.net_io_counters(pernic=False).bytes_recv, 2))
+
+
+def read_mem_free():
+    return mem_free_filter.filter(round((100 - psutil.virtual_memory().percent), 2))
+
+
+# ---------------------------------------------------------------------------
+# In this example, we demonstrate how System health and some simluated data
+# can be directed to two data center components (IoTCC and graphite) using Liota.
```
pmasrani: This comment does not seem relevant, we are sending data only to IoTCC in this case. In general, check all comments.

liota/lib/utilities/filters/windowed_filters/windowed_range_filter.py
```diff
@@ -0,0 +1,67 @@
+# -*- coding: utf-8 -*-
```
KohliDev: I believe in spite of adding this as the separate filter it should happen in the example code. In the UDM both filters(Window as well as value filter) should be applied, which will be more elaborative.

liota/lib/utilities/filters/windowed_filters/windowed_filter.py
```diff
@@ -0,0 +1,109 @@
...
+#  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR        #
+#  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF       #
+#  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS   #
+#  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN    #
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)    #
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF     #
+#  THE POSSIBILITY OF SUCH DAMAGE.                                            #
+# ----------------------------------------------------------------------------#
+
+from abc import abstractmethod
```
KohliDev: Add an example only for windows filter.

liota/lib/utilities/filters/windowed_filters/windowed_filter.py
```diff
@@ -0,0 +1,107 @@
...
+            if filtered_value is not None:
+                self.sample_passed = True  # At-least one sample has passed during this window
+            return filtered_value
+
+    def _set_next_window_time(self):
+        """
+        Sets next time-window.
+        :return: None
+        """
+        self.next_window_time = getUTCmillis() + (self.window_size_sec * 1000)
```
KohliDev: Agreed, we need to keep the trackof current`s windows end time and on basis of that get the new window end time.

liota/lib/utilities/filters/windowed_filters/windowed_range_filter.py
Venkat2811: With base classes `__init()__` having different signature, it is not possible to use `super()` in derived class. [Reference](http://stackoverflow.com/questions/16855087/how-does-multiple-inheritance-work-with-the-super-and-different-init-arg)
- One approach is to design base classes for multiple inheritance. [Reference] (http://stackoverflow.com/questions/34884567/python-multiple-inheritance-passing-arguments-to-constructors-using-super) (Possible use of kwargs)
- Other way is to use `BaseClass.__init__()`

So, I thought of using this approach.  Your opinion on this please ?

liota/lib/utilities/filters/range_filter.py
```diff
@@ -0,0 +1,181 @@
...
+    GREATER_THAN = 10
+    AT_LEAST = 11
+
+
+class RangeFilter(Filter):
+    """
+    A simple lightweight filter, that filters values based on the specified filter type (range).
+    """
+
+    def __init__(self, filter_type=None, lower_bound=None, upper_bound=None):
```
Venkat2811: Agreed. Default value is not necessary. Will change it.

liota/lib/utilities/filters/windowed_filters/windowed_filter.py
Venkat2811: Agreed.  Yes, it's a bug :)

liota/lib/utilities/filters/windowed_filters/windowed_range_filter.py
Venkat2811: We'll discuss it

Venkat2811:
Sure @KohliDev, that's the plan.

PR #53, opened 2016-12-12
Reused utility method

ApurvaH:


liota/lib/utilities/utility.py
```diff
@@ -153,3 +154,29 @@ def _find_path(self):
 
     def get_liota_fullpath(self):
         return LiotaConfigPath.path_liota_config
+
+def read_liota_config(section, name, data_type):
```
pmasrani: Passing of data_type as 0/1 values does not look good and seems confusing. If only collector_thread_pool_size is integer, then let this function return all properties as String, and let the caller convert it to integer. For example,

collect_thread_pool_size = int(read_liota_config('CORE_CFG','collect_thread_pool_size',))

PR #65, opened 2016-12-20
Converted psutil to linux_metrics

ApurvaH:


examples/dell5k_edge_system_iotcc.py
```diff
@@ -56,29 +55,35 @@
...
 
 def read_cpu_procs():
-    cnt = 0
-    procs = psutil.pids()
-    for i in procs[:]:
-        p = psutil.Process(i)
-        if p.status() == 'running':
-            cnt += 1
-    return cnt
+    from linux_metrics import cpu_stat
```
pmasrani: We can have all the imports at the top of the file, instead of in the function call. It is more convenient that way.

Issue #62, opened 2016-12-16
De-coupling IoTCC & HelixProtocol from DCC_comms layer


KohliDev:
De-coupling IoTCC & HelixProtocol from DCC_comms layer so that we get complete plug-play architecture for integrating with other DCC_Comms.

PR #67, opened 2016-12-23
Fixing the bug to decouple IoTCC & HelixProtocol from DCC_Comms layer

KohliDev:
Fix for the #62


liota/lib/transports/web_socket.py
```diff
@@ -120,18 +122,9 @@ def send(self, msg):
...
                 self.close()
-                sys.exit(0)
+                #sys.exit(0)
 
-    def next_id(self):
-        self.counter = (self.counter + 1) & 0xffffff
-        # Enforce even IDs
-        return self.counter * 2
-
-    def close(self):
```
Venkat2811: I think `close()` is deleted by mistake

liota/dcc_comms/mqtt_dcc_comms.py
```diff
@@ -124,17 +125,29 @@ def _disconnect(self):
         """
         self.client.disconnect()
 
-    def subscribe(self, msg_attr=None):
+    def receive(self,msg_attr=None):
```
Venkat2811: In-case of of Non-Project ICE auto-generation subscribe case, we should allow DCC to pass a callback method to receive method.  If that is None, we should make `receive_message`as the callback method.  So, method should be like:
```
    def receive(self, msg_attr=None, callback=None):
        if msg_attr:
            self.client.subscribe(msg_attr.sub_topic, msg_attr.sub_qos, msg_attr.sub_callback)
        else:
            self.msg_attr.sub_callback = callback if not None else self.receive_message
            self.client.subscribe(self.msg_attr.sub_topic, self.msg_attr.sub_qos, self.msg_attr.sub_callback)
```

In AMQP, I'm creating `AmqpPublishMessagingAttributes` and `AmqpSubscribeMessagingAttributes` as separate classes as there many params unlike `MqttMessagingAttributes`.  So, `receive()` in that case will take 3 args.  What I'm suggesting is AbstractDccComms's receive should be changed to 
```
@abstractmethod
def receive(**kwargs)
```

Also,`Mqtt README` should be updated.

liota/lib/transports/web_socket.py
```diff
@@ -76,60 +76,52 @@ def WebSocketConnection(self, host, verify_cert=True, CERTPATH="/etc/liota/cert"
             if self.ws is None:
                 raise (IOError("Couldn't verify host certificate"))
 
-    def run(self):
+    def receive(self, queue):
```
Venkat2811: I agree that there is no `receive()` in `WebSocket transports` like mqtt `subscribe()`.  But from code structure consistency POV, my suggestion is to keep it `WebSocket DccComms`.

KohliDev:
Code Changes to support MQTT and Websocket as DCC_Comms protocol for IoTCC.

examples/mqtt/dcc_comms/iotcc/dell5k_edge_system_iotcc_mqtt.py
```diff
@@ -0,0 +1,192 @@
...
+    # this object encapsulates the formats and protocols necessary for the agent to interact with the dcc
+    # UID/PASS login for now.
+
+    #  Creating EdgeSystem
+    edge_system = Dell5KEdgeSystem(config['EdgeSystemName'])
+    #  Encapsulates Identity
+    identity = Identity(root_ca_cert=None, username=config['broker_username'], password=config['broker_password'],
+                        cert_file=None, key_file=None)
+
+    iotcc = IotControlCenter(config['broker_username'], config['broker_password'],
```
Venkat2811: pass this way instead ?

`iotcc = IotControlCenter(identity.username, identity.password)`

KohliDev: Agreed. It will require change in all the examples, can be taken up in next pull request. If you are open, you can take it next pull request.

examples/dell5k_edge_system_iotcc.py
```diff
@@ -80,15 +80,15 @@ def read_mem_free():
 
     
 #---------------------------------------------------------------------------
-# In this example, we demonstrate how System health and some simluated data
-# can be directed to two data center components (IoTCC and graphite) using Liota.
+# In this example, we demonstrate how System health and some simulated data
+# can be directed to two data center components IoTCC using Liota.
```
pmasrani: two?

examples/dell5k_edge_system_iotcc.py
```diff
@@ -80,15 +80,15 @@ def read_mem_free():
```
pmasrani: Should we rename the file to indicate it is websocket based?

examples/mqtt/dcc_comms/iotcc/dell5k_edge_system_iotcc_mqtt.py
```diff
@@ -0,0 +1,192 @@
+# -*- coding: utf-8 -*-
```
pmasrani: Shouldn't the two examples based on websocket and MQTT be in the same location?

liota/dccs/iotcc.py
```diff
@@ -109,31 +111,25 @@ def on_receive_safe(msg):
...
                             else:
                                 log.info("Waiting for resource creation")
-                                time.sleep(5)
-                                self.con.send(
-                                    self._registration(self.con.next_id(), entity_obj.entity_id, entity_obj.name,
-                                                       entity_obj.entity_type))
+                                self.comms.send(json.dumps(
+                                    self._registration(self.next_id(), entity_obj.entity_id, entity_obj.name,
+                                                       entity_obj.entity_type)))
+                                on_receive_safe(self.recv_msg_queue.get())
```
pmasrani: I believe your assumption is that at any point in time, there is only one element in the queue, which happens to be the response for the pending request. This need not be true. For example, if one loads 10 packages each registering a different device, all requests will be from parallel threads and the queue will end up having responses for all. In such a case, this get() call will not be successful.

liota/lib/protocols/helix_protocol.py
```diff
@@ -31,7 +31,8 @@
 # ----------------------------------------------------------------------------#
 
 import logging
-
+import json
+import time
 
 PROTOCOL_VERSION = "2.7"
```
pmasrani: Need to change protocol version at both ends, so that older clients know it is a failure due to protocol mismatch

liota/dccs/iotcc.py
```diff
@@ -75,21 +77,21 @@ def on_receive_safe(msg):
             try:
```
pmasrani: Should we rename this method?

liota/dccs/iotcc.py
```diff
@@ -75,21 +77,21 @@ def on_receive_safe(msg):
...
-        thread = threading.Thread(target=self.con.run)
-        self.con.on_receive = on_receive_safe
+                    return True
+                else:
+                    log.debug("Processed msg: {0}".format(json_msg["type"]))
+                    on_receive_safe(self.recv_msg_queue.get())
+            except Exception as error:
+                log.error("HelixProtocolException: " + repr(error))
+
+        thread = threading.Thread(target=self.comms.receive)
```
pmasrani: Generally, underlying protocol libraries (Paho MQTT or Websocket-client) run a listener thread, when we add a receive method. Do we really need this thread in IOTCC now? We can await the execution till a message is received, i..e, till received queue is empty?

liota/lib/protocols/helix_protocol.py
```diff
@@ -53,7 +54,7 @@ def require_field(container, field):
 
 class HelixProtocol:
 
-    """ 
+    """
```
pmasrani: Update comments as well.

liota/lib/protocols/helix_protocol.py
```diff
@@ -128,125 +125,45 @@ def is_active(self):
...
 
 
-class HandshakeAwaitingState(State):
-
+class HandshakeRequestedState(State):
     def __init__(self, previous, proto=None):
         State.__init__(self, previous, proto)
-        self.name = "HandshakeAwaitingState"
+        self.name = "HandshakeRequestedState"
+        # require_field(msg, "transactionID")
```
pmasrani: remove commented code.

liota/lib/protocols/helix_protocol.py
```diff
@@ -128,125 +125,45 @@ def is_active(self):
         return self == self.proto.state
 
 
-class HandshakeAwaitingState(State):
-
+class HandshakeRequestedState(State):
```
pmasrani: Can you please elaborate why do we need this state?

liota/lib/transports/web_socket.py
```diff
@@ -52,14 +54,12 @@ def __init__(self, url):
     def connect_soc(self):
         try:
             self.WebSocketConnection(self.url, False)
-            log.info("Connection Successful")
         except Exception:
             log.exception("WebSocket exception, please check the WebSocket address and try again.")
-            sys.exit(0)
+            # sys.exit(0)
```
pmasrani: Remove commented code

liota/lib/transports/web_socket.py
pmasrani: I am not sure if i understand this. Websockets do have on_message(), which i think is similar to subscribe(). Am i missing something?

examples/dell5k_edge_system_iotcc.py
KohliDev: Removed.

examples/mqtt/dcc_comms/iotcc/dell5k_edge_system_iotcc_mqtt.py
KohliDev: This is a user program so we are keeping them in the MQTT folder for better accessibility. Packages are placed together.

examples/mqtt/dcc_comms/iotcc/dell5k_edge_system_iotcc_mqtt.py
KohliDev: Have to mark it as an issue.

liota/dcc_comms/mqtt_dcc_comms.py
KohliDev: The upper review comment is incorporated in the review request. 
received(**kwargs) needs to be discussed in detail.

liota/dccs/iotcc.py
KohliDev: renamed to on_response

liota/dccs/iotcc.py
KohliDev: We need to change Websocket client to WebSocket app to achieve his functionality, marked as an issue.

liota/dccs/iotcc.py
KohliDev: Marked as an issue as this bug is pat of the existing code.

liota/lib/protocols/helix_protocol.py
KohliDev: Changed:

2.8

liota/lib/protocols/helix_protocol.py
KohliDev: Updated

liota/lib/protocols/helix_protocol.py
KohliDev: To get hello message response and transaction id used for further communication in case of WebSocket

liota/lib/protocols/helix_protocol.py
KohliDev: Done

liota/lib/transports/web_socket.py
KohliDev: Done

liota/lib/transports/web_socket.py
KohliDev: This will be fixed when we move to WebSocketApp implementation marked as an issue.

liota/dcc_comms/mqtt_dcc_comms.py
Venkat2811: If auto-generated topic is used for subscription, users/dccs should ensure that message is dequeued from `userdata` queue. To avoid this, the below change has to be made.
```
def receive(self, msg_attr=None, callback=None):
        if msg_attr:
            self.client.subscribe(msg_attr.sub_topic, msg_attr.sub_qos, msg_attr.sub_callback)
        else:
            self.msg_attr.sub_callback = callback if callback else self.receive_message
            self.client.subscribe(self.msg_attr.sub_topic, self.msg_attr.sub_qos, self.msg_attr.sub_callback)
```

`receive(**kwargs)` will not be needed for now.  But, abstracted method should be changed to `receive(msg_attr=None, callback=None)` to overcome the above mentioned situation

liota/dcc_comms/mqtt_dcc_comms.py
```diff
@@ -124,17 +125,32 @@ def _disconnect(self):
...
-    def subscribe(self, msg_attr=None):
+    def receive(self, msg_attr=None):
         """
         Subscribes to a topic with specified QoS and callback.
+        Set call back to receive_message method if no callback method is passed by user.
 
         :param msg_attr: MqttMessagingAttributes Object
         :return:
         """
+        if msg_attr is not None and hasattr(msg_attr.sub_callback):
```
Venkat2811: `hasattr` check is redundant as `MqttMessagingAttributes` object will have it.

This way is more pythonic :)

```
callback = msg_attr.sub_callback if msg_attr else self.receive_message
```

liota/dcc_comms/mqtt_dcc_comms.py
```diff
@@ -124,17 +125,32 @@ def _disconnect(self):
         """
         self.client.disconnect()
 
-    def subscribe(self, msg_attr=None):
+    def receive(self, msg_attr=None):
```
Venkat2811: Please refer comment thread in outdated revision.

Venkat2811: As per our discussion, if auto-generated topic is used for subscription, users/dccs should wait on userdata queue and get the message.

KohliDev: As discussed, in case of auto-generation callback method will be enforced to be used.

liota/dcc_comms/mqtt_dcc_comms.py
KohliDev: Changed

liota/lib/protocols/helix_protocol.py
```diff
@@ -53,40 +54,23 @@ def require_field(container, field):
...
-    <-- connection_verified
+    <-- Hello
 
     handshake_verified
     ===================================================
-    for all output resource kinds simultaneously do
-        --> create_resource_kind_request
-        <-- create_resource_kind_response
+    --> connection_request
+    <-- create_response
```
KohliDev: Changed

liota/dcc_comms/mqtt_dcc_comms.py
KohliDev: I`ll update the README after the check-in.

packages/examples/mqtt/iotcc/iotcc_mqtt_edge_system_stats.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class PackageClass(LiotaPackage):
+
+    def run(self, registry):
+        import copy
+        from liota.entities.metrics.metric import Metric
+
+        # Acquire resources from registry
+        iotcc_edge_system = copy.copy(registry.get("iotcc_edge_system_mqtt"))
```
pmasrani: It does not seem to be a good idea to create a whole new package with the same metrics just because the underlying iotcc basic package is using a different protocol channel. Can we make this package (iotcc_edge_system_stats) independent of the underlying iotcc package?

KohliDev: As discussed, this is an application code, and it has been kept with the different name to have uniqueness during loading and making an entry in the registry. We cannot make iotcc_edge_system_stats independent of underlying iotcc package as it gets registered here. Probably, package manager might be required to re-designed.

PR #70, opened 2017-01-11
Setup branch

ApurvaH:
Modified setup.py


pmasrani:
This will get fixed in a more generic change: https://github.com/vmware/liota/pull/74

PR #74, opened 2017-01-27
Fixing file locations for more POSIX like compliance

warthog9:
More, or less, /etc is intended for configuration files, things like the examples, and packages, are (usually) expected to live in /usr/share/docs and /usr/share respectively.

This is a quick patch to shift things around a little bit so they are in more expected file locations, as well as coming more into line with what pip is expecting to do (I.E. it's not really intended, by their own admission, to install things into places like /etc).

Also went ahead and wrote up a quick file walker so that you don't need to list out the files explicitly for inclusion into the package from the source tree, which should make things in setup.py a little cleaner overall.


README.md
```diff
@@ -60,35 +60,35 @@ liota.conf provides path to find out various configuration & log files. When ini
 * Looks in the current working directory '.'
 * User's home directory '~'
 * A LIOTA_CONF environment variable
-* Finally the default location for every installation: /etc/liota/conf.
+* Finally the default location for every installation: /etc/liota/. (note this will need to be copied from the system doc directory, typically /usr/share/doc/liota-&lt;version&gt;/config/)
```
KohliDev: If I get this change correct, user have to manually copy the required files from system doc directory to "/etc/liota"

KohliDev:
@warthog9 Thanks for this pull request. I agree with you things like packages and example are not expected to be in /etc/liota folder and even pip is not expected to do so. We wanted to have single directory that will hold all the required  liota files making it easy for users. I ran installer i.e. setup.py from your code in my Ubuntu machine, I`m getting the files copied to the following location:

```
copying BSD_LICENSE.txt -> build/bdist.linux-x86_64/egg/share/doc/liota-0.2
copying BSD_NOTICE.txt -> build/bdist.linux-x86_64/egg/share/doc/liota-0.2
creating build/bdist.linux-x86_64/egg/share/doc/liota-0.2/examples
copying examples/model_simulated/graphite_bike_simulated.py -> build/bdist.linux-x86_64/egg/share/doc/liota-0.2/examples
copying examples/model_simulated/README.md -> build/bdist.linux-x86_64/egg/share/doc/liota-0.2/examples
copying examples/model_simulated/graphite_thermistor_simulated.py -> build/bdist.linux-x86_64/egg/share/doc/liota-0.2/examples
copying examples/model_simulated/iotcc_bike_simulated.py -> build/bdist.linux-x86_64/egg/share/doc/liota-0.2/examples
copying examples/model_simulated/iotcc_thermistor_simulated.py -> build/bdist.linux-x86_64/egg/share/doc/liota-0.2/examples
creating build/bdist.linux-x86_64/egg/share/doc/liota-0.2/packages
 ```

The path seems to be variable for different machines as per the environment. Can you suggest some alternate folder which is fixed and where we can keep all the required file to keep it simple for the users.

Thanks for the file walker, setup.py looks better now.

File: README.md (hunk 1 above)
Comment (warthog9) [inline README.md]: Correct, if installed via pip, the user would need to copy the config file to /etc and (likely) need to create the log directory as well (though that's not explicitly expressed in that config message).

**Comment by warthog9** [comment]
So pip has some pretty serious deficiencies when it comes to overall packaging (some of which are intentional design choices on the part of pip), it's not entirely it's fault, as pip is really only intended to deal with library packages for python, and it's not really intended to replace the distro level package manager.  With that in mind here's my overall thoughts:

From the general development perspective (I.E. someone is just going to install this in a virtualenv, not intending to go to production, etc), it doesn't really matter where the files go.  I'd argue it's generally a good idea to follow convention that other python packages take, but that's about it.

From the "deployment" perspective, I.E. non-development, it's probably best to work on an actual distro package, since that's how most entities are going to want to consume liota.  It's not as distro agnostic, but it will be "more right" for that specific distro.  I think it's also possible to get setup.py to generate distro packages if the right pieces are put together (I'll double check that though).  But this means that things like the libraries and docs will tend to end up where the distro has decided they should live, which is not guaranteed to be the same across distros.

To that end some examples to for reference on how they handled what we are discussing:
    https://github.com/pycurl/pycurl/blob/master/setup.py#L533 - General package, stashes things into share/doc, etc.  Doesn't really have a stand alone app portion, so I'm only pointing this out as a generic example

    https://gitlab.com/mailman/mailman/blob/master/setup.py / https://gitlab.com/mailman/mailman/blob/master/MANIFEST.in - Mailman3 (which is a bit closer to what liota is) takes a slightly different tact, and puts everything it cares about into /usr/lib/mailman, including it's docs and such.  Downside to this is that they've had to go through a very extensive process of trying to make the installation simpler in the general sense, but it (without a distro package) is basically an exercise left to the user to figure out how to bolt things into place, which I can appreciate from the liota perspective is not desirable (wanting to simplify setup).  

I don't have a great suggestion for how to keep this simple for the users to access post installation.  The route I took was to (generally) follow what most distros are going to expect (/usr/share, /usr/share/docs, etc), but that does spread the pieces out over the filesystem a bit more.  Mailman3 obviously took the direction of putting it all under /usr/lib/mailman but needs a more extensive setup script (which a simple one for liota could be added easily enough to copy the config files into place and make sure the directories are setup).  It's not a bad option, all the pieces are then in one place, but the setup is a little more complicated.  The only (according to POSIX anyway), place to stick everything so it's entirely self contained would be to put things into /opt, but pip isn't exactly geared for that either and it'll cause other problems, and is obviously not optimal.  From and end user perspective, distro packages are likely the best, as most folks likely won't want to `pip install` if they aren't doing development on liota itself.

I'm going to check with some other python folks, see if they have any better suggestions beyond what I've got listed above - it's possible I'm missing something else that I'm unaware of.  I'm also going to see if I can get the pieces needed to build some distro packages added to this, as that might be helpful as well in the long run.  Beyond that I don't have strong views on putting things into /usr/share/* or into /usr/lib is better or more right, but I put them both out there for your consideration as places to put things vs. /etc

(also of note what machine / distro / version are you using where /usr/share/liota and /usr/share/liota-<version> weren't created, so I can double check on that?)

**Comment by KohliDev** [comment]
I like the approach of mailman, if we can put everything under "/usr/lib/liota" including the conf and example files that will be good. We will remove "/etc/liota" folder completely and refer the json as well as conf files in liota codebase directly from "/usr/lib/liota". If this is the correct approach we can work on it or we`ll be very happy to accept the contribution from your side.

I tried the installation by forking your branch on MAC OS X El Captain(my dev machine) and Ubuntu 14.04 "/usr/share/liota" as well as "/usr/share/doc/liota" are not created in both the machines.

Please find the "setup.py" output below:

```
$ sudo python setup.py install                                                                                                                                                                   
running install                                                                                                                                                                                                                  
running bdist_egg                                                                                                                                                                                                                
running egg_info                                                                                                                                                                                                                 
writing requirements to liota.egg-info/requires.txt                                                                                                                                                                              
writing liota.egg-info/PKG-INFO                                                                                                                                                                                                  
writing top-level names to liota.egg-info/top_level.txt                                                                                                                                                                          
writing dependency_links to liota.egg-info/dependency_links.txt                                                                                                                                                                  
reading manifest file 'liota.egg-info/SOURCES.txt'                                                                                                                                                                               
reading manifest template 'MANIFEST.in'                                                                                                                                                                                          
warning: no files found matching 'example/sampleProp.conf'                                                                                                                                                                       
writing manifest file 'liota.egg-info/SOURCES.txt'                                                                                                                                                                               
installing library code to build/bdist.linux-x86_64/egg                                                                                                                                                                          
running install_lib                                                                                                                                                                                                              
running build_py                                                                                                                                                                                                                 
creating build/bdist.linux-x86_64/egg                                                                                                                                                                                            
creating build/bdist.linux-x86_64/egg/liota                                                                                                                                                                                      
creating build/bdist.linux-x86_64/egg/liota/entities                                                                                                                                                                             
copying build/lib.linux-x86_64-2.7/liota/entities/entity.py -> build/bdist.linux-x86_64/egg/liota/entities                                                                                                                       
creating build/bdist.linux-x86_64/egg/liota/entities/devices                                                                                                                                                                     
copying build/lib.linux-x86_64-2.7/liota/entities/devices/device.py -> build/bdist.linux-x86_64/egg/liota/entities/devices                                                                                                       
copying build/lib.linux-x86_64-2.7/liota/entities/devices/thermistor_simulated.py -> build/bdist.linux-x86_64/egg/liota/entities/devices                                                                                         
copying build/lib.linux-x86_64-2.7/liota/entities/devices/bike_simulated.py -> build/bdist.linux-x86_64/egg/liota/entities/devices                                                                                               
copying build/lib.linux-x86_64-2.7/liota/entities/devices/simulated_device.py -> build/bdist.linux-x86_64/egg/liota/entities/devices
copying build/lib.linux-x86_64-2.7/liota/entities/devices/__init__.py -> build/bdist.linux-x86_64/egg/liota/entities/devices
creating build/bdist.linux-x86_64/egg/liota/entities/edge_systems
copying build/lib.linux-x86_64-2.7/liota/entities/edge_systems/dell5k_edge_system.py -> build/bdist.linux-x86_64/egg/liota/entities/edge_systems
copying build/lib.linux-x86_64-2.7/liota/entities/edge_systems/__init__.py -> build/bdist.linux-x86_64/egg/liota/entities/edge_systems
copying build/lib.linux-x86_64-2.7/liota/entities/edge_systems/simulated_edge_system.py -> build/bdist.linux-x86_64/egg/liota/entities/edge_systems
copying build/lib.linux-x86_64-2.7/liota/entities/edge_systems/dk300_edge_system.py -> build/bdist.linux-x86_64/egg/liota/entities/edge_systems
copying build/lib.linux-x86_64-2.7/liota/entities/edge_systems/edge_system.py -> build/bdist.linux-x86_64/egg/liota/entities/edge_systems
copying build/lib.linux-x86_64-2.7/liota/entities/registered_entity.py -> build/bdist.linux-x86_64/egg/liota/entities
creating build/bdist.linux-x86_64/egg/liota/entities/metrics
copying build/lib.linux-x86_64-2.7/liota/entities/metrics/registered_metric.py -> build/bdist.linux-x86_64/egg/liota/entities/metrics
copying build/lib.linux-x86_64-2.7/liota/entities/metrics/metric.py -> build/bdist.linux-x86_64/egg/liota/entities/metrics
copying build/lib.linux-x86_64-2.7/liota/entities/metrics/__init__.py -> build/bdist.linux-x86_64/egg/liota/entities/metrics
copying build/lib.linux-x86_64-2.7/liota/entities/__init__.py -> build/bdist.linux-x86_64/egg/liota/entities
creating build/bdist.linux-x86_64/egg/liota/lib
creating build/bdist.linux-x86_64/egg/liota/lib/identity
copying build/lib.linux-x86_64-2.7/liota/lib/identity/identity.py -> build/bdist.linux-x86_64/egg/liota/lib/identity
copying build/lib.linux-x86_64-2.7/liota/lib/identity/tls_conf.py -> build/bdist.linux-x86_64/egg/liota/lib/identity
copying build/lib.linux-x86_64-2.7/liota/lib/identity/__init__.py -> build/bdist.linux-x86_64/egg/liota/lib/identity
creating build/bdist.linux-x86_64/egg/liota/lib/transports
copying build/lib.linux-x86_64-2.7/liota/lib/transports/socket.py -> build/bdist.linux-x86_64/egg/liota/lib/transports
copying build/lib.linux-x86_64-2.7/liota/lib/transports/web_socket.py -> build/bdist.linux-x86_64/egg/liota/lib/transports
copying build/lib.linux-x86_64-2.7/liota/lib/transports/__init__.py -> build/bdist.linux-x86_64/egg/liota/lib/transports
copying build/lib.linux-x86_64-2.7/liota/lib/transports/mqtt.py -> build/bdist.linux-x86_64/egg/liota/lib/transports
creating build/bdist.linux-x86_64/egg/liota/lib/protocols
copying build/lib.linux-x86_64-2.7/liota/lib/protocols/helix_protocol.py -> build/bdist.linux-x86_64/egg/liota/lib/protocols
copying build/lib.linux-x86_64-2.7/liota/lib/protocols/__init__.py -> build/bdist.linux-x86_64/egg/liota/lib/protocols
copying build/lib.linux-x86_64-2.7/liota/lib/__init__.py -> build/bdist.linux-x86_64/egg/liota/lib
creating build/bdist.linux-x86_64/egg/liota/lib/utilities
creating build/bdist.linux-x86_64/egg/liota/lib/utilities/filters
copying build/lib.linux-x86_64-2.7/liota/lib/utilities/filters/filter.py -> build/bdist.linux-x86_64/egg/liota/lib/utilities/filters
creating build/bdist.linux-x86_64/egg/liota/lib/utilities/filters/windowing_scheme
copying build/lib.linux-x86_64-2.7/liota/lib/utilities/filters/windowing_scheme/windowing_scheme.py -> build/bdist.linux-x86_64/egg/liota/lib/utilities/filters/windowing_scheme
copying build/lib.linux-x86_64-2.7/liota/lib/utilities/filters/windowing_scheme/__init__.py -> build/bdist.linux-x86_64/egg/liota/lib/utilities/filters/windowing_scheme
copying build/lib.linux-x86_64-2.7/liota/lib/utilities/filters/__init__.py -> build/bdist.linux-x86_64/egg/liota/lib/utilities/filters
copying build/lib.linux-x86_64-2.7/liota/lib/utilities/filters/range_filter.py -> build/bdist.linux-x86_64/egg/liota/lib/utilities/filters
copying build/lib.linux-x86_64-2.7/liota/lib/utilities/__init__.py -> build/bdist.linux-x86_64/egg/liota/lib/utilities
copying build/lib.linux-x86_64-2.7/liota/lib/utilities/utility.py -> build/bdist.linux-x86_64/egg/liota/lib/utilities
copying build/lib.linux-x86_64-2.7/liota/lib/utilities/si_unit.py -> build/bdist.linux-x86_64/egg/liota/lib/utilities
creating build/bdist.linux-x86_64/egg/liota/dccs
copying build/lib.linux-x86_64-2.7/liota/dccs/dcc.py -> build/bdist.linux-x86_64/egg/liota/dccs
copying build/lib.linux-x86_64-2.7/liota/dccs/graphite.py -> build/bdist.linux-x86_64/egg/liota/dccs
copying build/lib.linux-x86_64-2.7/liota/dccs/iotcc.py -> build/bdist.linux-x86_64/egg/liota/dccs
copying build/lib.linux-x86_64-2.7/liota/dccs/__init__.py -> build/bdist.linux-x86_64/egg/liota/dccs
creating build/bdist.linux-x86_64/egg/liota/core
copying build/lib.linux-x86_64-2.7/liota/core/package_manager.py -> build/bdist.linux-x86_64/egg/liota/core
copying build/lib.linux-x86_64-2.7/liota/core/__init__.py -> build/bdist.linux-x86_64/egg/liota/core
copying build/lib.linux-x86_64-2.7/liota/core/metric_handler.py -> build/bdist.linux-x86_64/egg/liota/core
creating build/bdist.linux-x86_64/egg/liota/device_comms
copying build/lib.linux-x86_64-2.7/liota/device_comms/device_comms.py -> build/bdist.linux-x86_64/egg/liota/device_comms
copying build/lib.linux-x86_64-2.7/liota/device_comms/__init__.py -> build/bdist.linux-x86_64/egg/liota/device_comms
copying build/lib.linux-x86_64-2.7/liota/device_comms/mqtt_device_comms.py -> build/bdist.linux-x86_64/egg/liota/device_comms
copying build/lib.linux-x86_64-2.7/liota/__init__.py -> build/bdist.linux-x86_64/egg/liota
creating build/bdist.linux-x86_64/egg/liota/dcc_comms
copying build/lib.linux-x86_64-2.7/liota/dcc_comms/dcc_comms.py -> build/bdist.linux-x86_64/egg/liota/dcc_comms
copying build/lib.linux-x86_64-2.7/liota/dcc_comms/websocket_dcc_comms.py -> build/bdist.linux-x86_64/egg/liota/dcc_comms
copying build/lib.linux-x86_64-2.7/liota/dcc_comms/mqtt_dcc_comms.py -> build/bdist.linux-x86_64/egg/liota/dcc_comms
copying build/lib.linux-x86_64-2.7/liota/dcc_comms/__init__.py -> build/bdist.linux-x86_64/egg/liota/dcc_comms
copying build/lib.linux-x86_64-2.7/liota/dcc_comms/socket_comms.py -> build/bdist.linux-x86_64/egg/liota/dcc_comms
creating build/bdist.linux-x86_64/egg/liota/actions
copying build/lib.linux-x86_64-2.7/liota/actions/__init__.py -> build/bdist.linux-x86_64/egg/liota/actions
creating build/bdist.linux-x86_64/egg/liota/alerts
copying build/lib.linux-x86_64-2.7/liota/alerts/__init__.py -> build/bdist.linux-x86_64/egg/liota/alerts
byte-compiling build/bdist.linux-x86_64/egg/liota/entities/entity.py to entity.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/entities/devices/device.py to device.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/entities/devices/thermistor_simulated.py to thermistor_simulated.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/entities/devices/bike_simulated.py to bike_simulated.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/entities/devices/simulated_device.py to simulated_device.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/entities/devices/__init__.py to __init__.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/entities/edge_systems/dell5k_edge_system.py to dell5k_edge_system.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/entities/edge_systems/__init__.py to __init__.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/entities/edge_systems/simulated_edge_system.py to simulated_edge_system.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/entities/edge_systems/dk300_edge_system.py to dk300_edge_system.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/entities/edge_systems/edge_system.py to edge_system.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/entities/registered_entity.py to registered_entity.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/entities/metrics/registered_metric.py to registered_metric.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/entities/metrics/metric.py to metric.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/entities/metrics/__init__.py to __init__.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/entities/__init__.py to __init__.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/lib/identity/identity.py to identity.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/lib/identity/tls_conf.py to tls_conf.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/lib/identity/__init__.py to __init__.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/lib/transports/socket.py to socket.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/lib/transports/web_socket.py to web_socket.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/lib/transports/__init__.py to __init__.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/lib/transports/mqtt.py to mqtt.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/lib/protocols/helix_protocol.py to helix_protocol.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/lib/protocols/__init__.py to __init__.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/lib/__init__.py to __init__.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/lib/utilities/filters/filter.py to filter.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/lib/utilities/filters/windowing_scheme/windowing_scheme.py to windowing_scheme.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/lib/utilities/filters/windowing_scheme/__init__.py to __init__.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/lib/utilities/filters/__init__.py to __init__.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/lib/utilities/filters/range_filter.py to range_filter.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/lib/utilities/__init__.py to __init__.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/lib/utilities/utility.py to utility.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/lib/utilities/si_unit.py to si_unit.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/dccs/dcc.py to dcc.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/dccs/graphite.py to graphite.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/dccs/iotcc.py to iotcc.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/dccs/__init__.py to __init__.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/core/package_manager.py to package_manager.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/core/__init__.py to __init__.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/core/metric_handler.py to metric_handler.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/device_comms/device_comms.py to device_comms.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/device_comms/__init__.py to __init__.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/device_comms/mqtt_device_comms.py to mqtt_device_comms.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/__init__.py to __init__.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/dcc_comms/dcc_comms.py to dcc_comms.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/dcc_comms/websocket_dcc_comms.py to websocket_dcc_comms.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/dcc_comms/mqtt_dcc_comms.py to mqtt_dcc_comms.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/dcc_comms/__init__.py to __init__.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/dcc_comms/socket_comms.py to socket_comms.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/actions/__init__.py to __init__.pyc
byte-compiling build/bdist.linux-x86_64/egg/liota/alerts/__init__.py to __init__.pyc
installing package data to build/bdist.linux-x86_64/egg
running install_data
creating build/bdist.linux-x86_64/egg/share
creating build/bdist.linux-x86_64/egg/share/doc
creating build/bdist.linux-x86_64/egg/share/doc/liota-0.2
copying BSD_LICENSE.txt -> build/bdist.linux-x86_64/egg/share/doc/liota-0.2
copying BSD_NOTICE.txt -> build/bdist.linux-x86_64/egg/share/doc/liota-0.2
creating build/bdist.linux-x86_64/egg/share/doc/liota-0.2/examples
copying examples/model_simulated/graphite_bike_simulated.py -> build/bdist.linux-x86_64/egg/share/doc/liota-0.2/examples
copying examples/model_simulated/README.md -> build/bdist.linux-x86_64/egg/share/doc/liota-0.2/examples
copying examples/model_simulated/graphite_thermistor_simulated.py -> build/bdist.linux-x86_64/egg/share/doc/liota-0.2/examples
copying examples/model_simulated/iotcc_bike_simulated.py -> build/bdist.linux-x86_64/egg/share/doc/liota-0.2/examples
copying examples/model_simulated/iotcc_thermistor_simulated.py -> build/bdist.linux-x86_64/egg/share/doc/liota-0.2/examples
creating build/bdist.linux-x86_64/egg/share/doc/liota-0.2/packages
copying packages/edge_systems/dell5k/edge_system.py -> build/bdist.linux-x86_64/egg/share/doc/liota-0.2/packages
creating build/bdist.linux-x86_64/egg/share/doc/liota-0.2/config
copying config/README.md -> build/bdist.linux-x86_64/egg/share/doc/liota-0.2/config
copying config/liota.conf -> build/bdist.linux-x86_64/egg/share/doc/liota-0.2/config
copying config/logging.json -> build/bdist.linux-x86_64/egg/share/doc/liota-0.2/config
creating build/bdist.linux-x86_64/egg/share/liota
creating build/bdist.linux-x86_64/egg/share/liota/packages
copying packages/edge_systems/dell5k/edge_system.py -> build/bdist.linux-x86_64/egg/share/liota/packages
creating build/bdist.linux-x86_64/egg/EGG-INFO
copying liota.egg-info/PKG-INFO -> build/bdist.linux-x86_64/egg/EGG-INFO
copying liota.egg-info/SOURCES.txt -> build/bdist.linux-x86_64/egg/EGG-INFO
copying liota.egg-info/dependency_links.txt -> build/bdist.linux-x86_64/egg/EGG-INFO
copying liota.egg-info/requires.txt -> build/bdist.linux-x86_64/egg/EGG-INFO
copying liota.egg-info/top_level.txt -> build/bdist.linux-x86_64/egg/EGG-INFO
zip_safe flag not set; analyzing archive contents...
creating 'dist/liota-0.2-py2.7.egg' and adding 'build/bdist.linux-x86_64/egg' to it
removing 'build/bdist.linux-x86_64/egg' (and everything under it)
Processing liota-0.2-py2.7.egg
Removing /usr/local/lib/python2.7/dist-packages/liota-0.2-py2.7.egg
Copying liota-0.2-py2.7.egg to /usr/local/lib/python2.7/dist-packages
liota 0.2 is already the active version in easy-install.pth

Installed /usr/local/lib/python2.7/dist-packages/liota-0.2-py2.7.egg
Processing dependencies for liota==0.2
Searching for aenum==1.4.5
Best match: aenum 1.4.5
Processing aenum-1.4.5-py2.7.egg
aenum 1.4.5 is already the active version in easy-install.pth

Using /usr/local/lib/python2.7/dist-packages/aenum-1.4.5-py2.7.egg
Searching for paho-mqtt==1.2
Best match: paho-mqtt 1.2
Processing paho_mqtt-1.2-py2.7.egg
paho-mqtt 1.2 is already the active version in easy-install.pth

Using /usr/local/lib/python2.7/dist-packages/paho_mqtt-1.2-py2.7.egg
Searching for Pint==0.7.2
Best match: Pint 0.7.2
Adding Pint 0.7.2 to easy-install.pth file

Using /usr/local/lib/python2.7/dist-packages
Searching for linux-metrics==0.1.4
Best match: linux-metrics 0.1.4
Adding linux-metrics 0.1.4 to easy-install.pth file

Using /usr/local/lib/python2.7/dist-packages
Searching for websocket-client==0.37.0
Best match: websocket-client 0.37.0
Adding websocket-client 0.37.0 to easy-install.pth file

Using /usr/local/lib/python2.7/dist-packages
Searching for backports.ssl-match-hostname==3.5.0.1
Best match: backports.ssl-match-hostname 3.5.0.1
Adding backports.ssl-match-hostname 3.5.0.1 to easy-install.pth file

Using /usr/local/lib/python2.7/dist-packages
Searching for six==1.10.0
Best match: six 1.10.0
Adding six 1.10.0 to easy-install.pth file

Using /usr/local/lib/python2.7/dist-packages
Finished processing dependencies for liota==0.2

```

**Comment by KohliDev** [comment]
Hello @warthog9 

Would like to know if there any suggestions from Python folks on the directory structure? 
For time being I think we should be able to merge the tree structure code base (part of setup.py) if you can raise it in the separate pull request or edit the current one.

Thanks,
Kohli

**Comment by warthog9** [comment]
We will remove "/etc/liota" folder completely and refer the json as well as conf files in liota codebase 
    directly from "/usr/lib/liota". If this is the correct approach we can work on it or we`ll be very happy to 
    accept the contribution from your side.

I think the general consensus is that the config files should still end up under /etc.  I.E. anything that the user would expect to edit / change (I.E. config files) should end up under /etc.  Anything that's site-packages, ends up in site-packages (I.E. library kind of stuff).  Anything that's needed for running the website type stuff, ends up under /usr/lib/liota.

I'll double check this on OS X once I've got it re-worked.  It's possible the OS X has a pathing that I'm not expecting, which would (as always) complicate things slightly.

**Comment by warthog9** [comment]
Ok moved to /usr/lib now, I should double check what OS X would likely be expecting, and it's not clear what Windows (should this be installed on Windows) would "prefer" per-se.  Also realized I was accidentally putting the files in the egg itself, and not on the absolute path of the system - so that's fixed now as well.

**Comment by warthog9** [comment]
Fixed a bug in directory discovery, otherwise this is mostly just keeping it up to date with the current changes in liota for merging.

**Comment by KohliDev** [comment]
@warthog9, Thanks for the PR update, I`ll test the installation from your dev branch and let you know the results by this weekend.

**Comment by warthog9** [comment]
Any news?

**Comment by KohliDev** [comment]
@warthog9 Sorry for the delay. I tested out the code changes by cloning your branch.

Some questions & thoughts:

1. If "/usr/lib/liota" will always get created on all the Linux systems. We can change by default path of Liota from "/etc/liota" to "/usr/lib/liota" and in this case, the user doesn't need to copy the files manually by creating "/etc/liota" folder. I'm fine with this approach if it goes well with the python community.

2. I tried installing on Ubuntu system it worked fine but installation fails on mac, we officially don`t support MAC but the primary requirement is to support most of the Linux systems:

```
creating /usr/lib/liota
error: could not create '/usr/lib/liota': Operation not permitted
```

**Comment by warthog9** [comment]
1) So the problem here is that *nix type operating systems pretty much define that files will get put all over the filesystem in specific locations.  Configuration files are supposed to end up in /etc, user space libraries in /usr/lib, very core system libraries /lib, etc.  This thinking has been encapsulated in a lot of how these OSes work, and in particular with regards to security.  Things like SELinux will actively enforce the separation of these directories, which is one of the reasons I'm proposing moving the files out of /etc in the first place.  Not having these files, eventually, in the right places will make getting packages into the distros pretty hard, if not impossible.  /etc/liota will likely need to contain the configuration file for the process, and /usr/lib is where I was stashing everything else.  Considering that setup.py / pip aren't intended as generic package managers it complicates things beyond that, and the manual copy step is nigh unavoidable.  This *IS* fixable in a proper package, outside of setup.py / pip though, so things like .rpm and .deb files this can be fixed, which is what you'd expect a normal, and typical, deployment for non-developers to look like.  From an ease of use perspective I'd recommend focusing on getting that right in the packages vs. in the setup.py / pip, which the later is only going to really get used by developers.

2) Mac uses a slightly different hierarchy, let me double check where OS X is expecting those to go and add that in, just need to figure out how to detect Linux vs. OS X

**Comment by KohliDev** [comment]
1. Thanks for the explanation and I kind of agree with the first point. But the only way as of now to get Liota to multiple OS's is to get it installed through pip. And as we discussed previously we just want to avoid the pain of forcing the users to manually copy the files(We can look for some alternative in setup.py to automate the process of copying the conf files to /etc/liota folder). So, we kind of have to figure out the way to either keep the conf files in /etc/liota and other stuff in /usr/lib/liota or alternative solution is to use one directory to keep all the stuff. We can't go by the way to manually ask the user to copy the files.

2. We can skip the MAC installation as of now and just keep the setup.py installation restricted to Linux systems.

**Comment by warthog9** [comment]
1) is the objection that there isn't a script, that after pip install lets say, that would do a "post install tasks" kind of deal, such as creating the directory and copying initial content to it.  Or is it that either a script, or manual processes, would need to be done at all?

If it's the former I can write a script quickly enough, but it obviously would still incur a step of `pip install liota && ./<script in directory>`.

If it's the later, you are pointing out a fundamental design choice with pip that it's not a general package manager, and doesn't want to be, but that you are pointing out is the only way to (currently) distribute liota, so I'm not sure how to move past that fundamental limitation in pip.

2) I'll take a look at it, it shouldn't be hard, just don't have a lot of Macs sitting around to test against.

**Comment by KohliDev** [comment]
pip install liota && ./<script in directory> looks like the only option available. Let`s have meeting with @gbollella and liota-dev members to discuss this change and then we can merge the pull request.

**Comment by KohliDev** [comment]
Hello @warthog9 ,
I`m yet to have the discussion with @GregBollella. Sorry for the delay, will update you by next week.

Thanks,
Kohli

**Comment by KohliDev** [comment]
@warthog9 I discussed this PR with @GregBollella during Radio. One suggestion which I have is we can also bundle the examples folder separately and allow the user to download them separately as tar folder from github repo. 
But in this approach also we`ll have to copy the conf file manually using add-on script as in the current state we are heavily dependent on the configuration files to run Liota.

What will be the correct path in Linux to keep the liota conf files and example folder, I strongly believe "/etc/liota" is not the correct path? Your thoughts

**Comment by KohliDev** [comment]
@warthog9 Let me know if we can have a meeting to discuss the solution for this PR.

**Comment by KohliDev** [comment]
If we don`t have any alternate mechanism you can progress on the installer @warthog9.
We`ll try and merge the pull request by next week.

File: README.md [hunk 2]
```diff
@@ -121,37 +121,37 @@ liota.conf provides path to find out various configuration & log files. When ini
...
 
 [LOG_PATH]
 log_path = /var/log/liota
 
 [UUID_PATH]
-uuid_path = /etc/liota/conf/uuid.ini
+uuid_path = /etc/liota/uuid.ini
 
 [IOTCC_PATH]
 dev_file_path = /etc/liota/conf/devs
```
KohliDev: Will it be fine to change IoTCC Path as well from "/etc/liota/conf" to "/etc/liota" as by default syswide path is changed to  '/etc/liota/' in utility.py?

post-install-setup.sh
```diff
@@ -0,0 +1,237 @@
...
+#	This will modify your system, should be run as root (it's going to 
+#	create directories and such in places that a normal user would not
+#	have access to, and you should always be cautious in running this,
+#	or any software as root, at the best of times.
+#
+# ----------------------------------------------------------------------------
+# Things that are being done / todo list
+#	- Check for the users we'll be using
+#		- Check the root user
+#		- Check the liota user
```
KohliDev: I believe we can only check for the root user as I`m not sure if the "liota" user and group will be allowed to be created on every Edge Gateway.

post-install-setup.sh
```diff
@@ -0,0 +1,237 @@
...
+#		- Check the liota group
+#	- Configuration files: /etc/liota
+#		- If /etc/liota doesn't exist, make it
+#		- Change the owner and group to what we are expecting
```
KohliDev: User can create and change the ownership manually as required by him. We can comment out the Liota user code and can use it later if we are allowed to create new "liota" user on the gateway. I can check and can get back by tomorrow.

post-install-setup.sh
```diff
@@ -0,0 +1,237 @@
...
+
+cp "${_LIOTA_CONFFROM}" "${_LIOTA_CONFTO}"
+
+if [[ "$?" != "0" ]]
+then
+	echo "ERROR: couldn't copy ${_LIOTA_CONFFROM} to ${_LIOTA_CONFTO}.  This is a fatal error."
+	exit
+fi
+
+chown "${_LIOTA_ROOTUSER}:${_LIOTA_USER}" "${_LIOTA_CONFTO}"
```
KohliDev: User should handle this process manually(chown) as we are not sure if we can create a new "liota" user on the IoT Gateway.

post-install-setup.sh
```diff
@@ -0,0 +1,237 @@
...
+fi
+
+
+#
+# Configuration directory
+#
+
+if [[ -e "${_LIOTA_ETCDIR}" ]]
+then
+	echo "ERROR: ${_LIOTA_ETCDIR} already exists, and assuming things are setup already or not but automatic running will not proceed. This is a fatal error."
```
KohliDev: But in Liota upgrade scenario if some new parameters are introduced in config files(new liota release) then I believe this script won`t be replacing the older pre-existing config files in the "/etc/liota" directory? Will the user have to manually copy them?

README.md
warthog9: Should be fine, I'll add that in

post-install-setup.sh
warthog9: Is liota really intended to be run as root? That would seem non-optimal to me.

I'm reluctant to create the liota user automatically in this helper script as it's intended to help with a dev setup, not be a a means for production deployment. Thus I'm assuming the dev will want to set up the user on their own, or in the worst case use root (which I would highly not recommend)

post-install-setup.sh
warthog9: So you don't want user permissions set?

post-install-setup.sh
KohliDev: Liota is not intended to run as root user, it will be running as a non-root user. I would go with the second approach i.e. inspite of we creating a Liota user in the script, dev should setup up own user and change the configuration file permissions.

post-install-setup.sh
KohliDev: I would prefer entire user creation and permissions steps for Liota user and group to be manually performed by dev rather than by this script.

post-install-setup.sh
warthog9: This is not a production deployment script, it's a dev environment helper. Even in a production deployment the package manager world *NEVER* replace the conf file with a new one. Distros like Debian can try to patch in new config info, but that has mixed results and usually requires manual intervention anyway.

If there's enough churn in the config file it basically will require manual user intervention to merge the respective changes, so no I'm not going to replace the config file if one exists, and a user would have to copy it in themselves

post-install-setup.sh
warthog9: Ok given the other conversation, should I comment this out, or leave it for now?

KohliDev: We should leave it for now.

warthog9:
Added a couple more things, and included copying the json logging config into place as well as making sure that the logging directory gets setup

post-install-setup.sh
```diff
@@ -0,0 +1,328 @@
+#!/bin/bash
```
KohliDev: Hello @warthog9, I`m facing issues in running this script:

1. After creating liota user:
```
$ ./post-install-setup.sh
Gives following error:
Liota: etc dir: /etc/liota
ERROR: /etc/liota already exists, and assuming things are setup already or not but automatic running will not proceed. This is a fatal error.
```

It only copies one file liota.conf but not logging.json.

2. Also, running the script with root user gives the following error:
```
 $ _LIOTA_ROOTUSER="root" ./post-install-setup.sh 
Liota: etc dir: /etc/liota
ERROR: There is a configuration file already existing at /etc/liota/liota.conf, not going to overwrite. This is a fatal error.
```

I tried deleting the "/etc/liota" and retrying above commands but it fails again. I believe there is a double check in the script one of them will have to be removed.

post-install-setup.sh
```diff
@@ -0,0 +1,328 @@
...
+
+chmod "${_LIOTA_CONFPERMS}" "${_LIOTA_CONFTO}"
+
+if [[ "$?" != "0" ]]
+then
+	echo "ERROR: couldn't set permissions to ${_LIOTA_CONFPERMS} on ${_LIOTA_CONFTO}. This is a fatal error."
+	exit
+fi
+
+if [[ -e "${_LIOTA_CONFTO}" ]]
```
KohliDev: This is required to be changed to _LIOTA_JSONTO

Issue #78, opened 2017-02-06
Add filter, windowing mechanism in Graphite_Edge_System example in packages


KohliDev:
@Venkat2811 

Even change eth0 to ens33 as the network interface.

Venkat2811:
Addressed in #79

PR #79, opened 2017-02-07
Filter with Graphite

Venkat2811:
Fix for #78
----------- 
 - eth0 > ens33
- log.info > log.debug
- `ValueError` > `TypeError`
- Applying Windowing scheme to `CPUUtil` instead of Network Util

### DIFF
--- a/examples/dell5k_edge_system_iotcc.py
+++ b/examples/dell5k_edge_system_iotcc.py
@@ -69,7 +69,7 @@ def read_disk_usage_stats():
 
 
 def read_network_bytes_received():
-    return round(net_stat.rx_tx_bytes('eth0')[0], 2)
+    return round(net_stat.rx_tx_bytes('ens33')[0], 2)
 
 
 def read_mem_free():
--- a/liota/lib/utilities/filters/range_filter.py
+++ b/liota/lib/utilities/filters/range_filter.py
@@ -92,7 +92,7 @@ class RangeFilter(Filter):
         """
         if filter_type not in Type:
             log.error("Unsupported Filter Type.")
-            raise ValueError("Unsupported Filter Type.")
+            raise TypeError("Unsupported Filter Type.")
 
         self.filter_type = filter_type
         self._validate(lower_bound, upper_bound)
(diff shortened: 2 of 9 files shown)


examples/dell5k_edge_system_iotcc.py
```diff
@@ -69,7 +69,7 @@ def read_disk_usage_stats():
 
 
 def read_network_bytes_received():
-    return round(net_stat.rx_tx_bytes('eth0')[0], 2)
+    return round(net_stat.rx_tx_bytes('ens33')[0], 2)
```
Venkat2811: Thanks for sharing reference, Piyush. `route | grep '^default' | grep -o '[^ ]*$'` command works with Linux OS.  I was looking for other solutions also.  We need to identify OS type and execute corresponding commands to get primary interface.

Since this is an example, can we get interface name from `sampleProp.conf` instead?

KohliDev: Thanks for pointing this out. We should surely take up the better approach but as this code is required urgently we should go ahead and push the code. Will open an issue for the same.

liota/lib/utilities/filters/range_filter.py
```diff
@@ -176,8 +176,8 @@ def filter(self, v):
             result = v
 
         if result is not None:
-            log.info("Value passed by filter : %s" % str(result))
+            log.debug("Value passed by filter : %s" % str(result))
```
KohliDev: Keep it as info logs, as of know

examples/dell5k_edge_system_iotcc.py
pmasrani: @Venkat2811 As we are using linux_metrics library for the examples, our code can anyway be run for Linux OS alone. Hence, I guess it should be fine to use Linux specific commands. When we introduce example for other OS (Windows), we can rework the example without linux_metrics library and linux specific command.

Venkat2811: Sure. #80 will use this approach.

PR #87, opened 2017-02-17
Filter unit test cases

Venkat2811:
Unit test cases for RangeFilter and WindowingScheme


tests/unit/lib/utilities/filters/test_filters.py
```diff
@@ -0,0 +1,185 @@
...
+        at_least_filter = RangeFilter(Type.AT_LEAST, self.lower_bound, self.upper_bound)
+        out_range_value = at_least_filter.filter(0)
+        in_range_value = at_least_filter.filter(30)
+        self.assertEquals(out_range_value, None)
+        self.assertEquals(in_range_value, 30)
+
+if __name__ == '__main__':
+    unittest.main(verbosity=1)
+
+
```
KohliDev: Remove the extra lines.

tests/unit/lib/utilities/filters/test_filters.py
```diff
@@ -0,0 +1,183 @@
...
+    def test_number_closed(self):
+        """If any filter applied on value which is not a number, value will be returned as it is"""
+        closed_filter = RangeFilter(Type.CLOSED, self.lower_bound, self.upper_bound)
+        value = closed_filter.filter("6")
+        self.assertEquals(value, "6")
+
+    def test_closed_filter(self):
+        """If CLOSED filter applied value will be accepted if: lower_bound <= value <= upper_bound"""
+        closed_filter = RangeFilter(Type.CLOSED, self.lower_bound, self.upper_bound)
+        value = closed_filter.filter(15)
```
pmasrani: Use another variable, like middle_value, and define it as (lower_bound+upper_bound)/2. That way even if somebody changes these values, test cases will not start failing.

tests/unit/lib/utilities/filters/test_filters.py
```diff
@@ -0,0 +1,183 @@
...
+        self.assertEquals(value, 15)
+
+    def test_open_filter(self):
+        """If OPEN filter applied value will be accepted if: lower_bound < value < upper_bound"""
+        open_filter = RangeFilter(Type.OPEN, self.lower_bound, self.upper_bound)
+        out_range_value = open_filter.filter(10)
```
pmasrani: Again, use the variable names.

tests/unit/lib/utilities/filters/test_filters.py
```diff
@@ -0,0 +1,183 @@
...
+        in_range_value = open_filter.filter(15)
+        self.assertEquals(out_range_value, None)
+        self.assertEquals(in_range_value, 15)
+
+    def test_closed_open_filter(self):
+        """If CLOSED_OPEN filter applied value will be accepted if: lower_bound <= value < upper_bound """
+        closed_open_filter = RangeFilter(Type.CLOSED_OPEN, self.lower_bound, self.upper_bound)
+        out_range_value = closed_open_filter.filter(20)
```
pmasrani: Ditto. Also might have to assert if the actual outside range values (like 5 and 25) are getting filtered or not.

tests/unit/lib/utilities/filters/test_filters.py
```diff
@@ -0,0 +1,183 @@
...
+        """If GREATER_THAN filter applied value will be accepted if: value > upper_bound"""
+        greater_than_filter = RangeFilter(Type.GREATER_THAN, self.lower_bound, self.upper_bound)
+        out_range_value = greater_than_filter.filter(20)
+        in_range_value = greater_than_filter.filter(30)
+        self.assertEquals(out_range_value, None)
+        self.assertEquals(in_range_value, 30)
+
+    def test_at_least(self):
+        """If AT_LEAST filter applied value will be accepted if: value >= upper_bound"""
+        at_least_filter = RangeFilter(Type.AT_LEAST, self.lower_bound, self.upper_bound)
```
pmasrani: In general, this are a lot of test cases and if someone modifies the numbers in future, all test cases might have to be fixed. I think better idea will be to define less_than_lower_bound= 5 lower_bound=10, middle_value = 15, upper_bound = 20, more_than_upper_bound=25 etc. and then write test cases to validate for all.

tests/unit/lib/utilities/filters/test_filters.py
Venkat2811: In this filter type, values will be filtered out if this condition is met: lower_bound <= value < upper_bound.  So, here lower bound itself is outside range.  Another test for outside range would be redundant.

tests/unit/lib/utilities/filters/test_filters.py
Venkat2811: Agreed.  Will use variables whose values are derived from lower & upper bound limit

KohliDev:
We need to integrate these test cases with Jenkins pipeline.

PR #90, opened 2017-02-26
SensorTag as Device

Venkat2811:
- `SensorTagCollector` with auto-reconnect mechanism as BLE exception occurs frequently

PR #91, opened 2017-03-02
add an example of subscribing to an mqtt device topic

tompscanlan:
Submitted for discussion.  I'm guessing this was the direction you were headed, right?

This is a package for listening to on an mqtt topic.  A device would publish to the topic and we receive the message.  The next step would be to pass this on to the DCC.

I'm adding this because I'd like to set up some demo devices and wanted to make sure this is what you were expecting that would look like.

### DIFF
--- a/liota/entities/devices/mqtt_subscription.py
+++ b/liota/entities/devices/mqtt_subscription.py
@@ -1,0 +1,91 @@
+# -*- coding: utf-8 -*-
+# ----------------------------------------------------------------------------#
+#  Copyright © 2017 VMware, Inc. All Rights Reserved.                         #
+#                                                                             #
+#  Licensed under the BSD 2-Clause License (the “License”); you may not use   #
+#  this file except in compliance with the License.                           #
+#                                                                             #
+#  The BSD 2-Clause License                                                   #
+#                                                                             #
+#  Redistribution and use in source and binary forms, with or without         #
+#  modification, are permitted provided that the following conditions are met:#
+#                                                                             #
+#  - Redistributions of source code must retain the above copyright notice,   #
+#      this list of conditions and the following disclaimer.                  #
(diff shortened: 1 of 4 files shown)


liota/entities/devices/mqtt_subscription.py
```diff
@@ -0,0 +1,91 @@
...
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)    #
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF     #
+#  THE POSSIBILITY OF SUCH DAMAGE.                                            #
+# ----------------------------------------------------------------------------#
+
+import logging
+
+import pint
+
+from liota.device_comms.mqtt_device_comms import MqttDeviceComms
```
KohliDev: @tompscanlan The device entity should remain independent of the device_comms. 
Basically, the con object can be passed to it for collecting the message over device_comms and parsing them in required device format. 

Please refer DCC entity, device should be designed in a similar way.

https://github.com/vmware/liota/blob/master/liota/dccs/iotcc.py

Also, refer this pull request for more details #90

packages/examples/mqtt_device.py
```diff
@@ -0,0 +1,61 @@
...
+# ----------------------------------------------------------------------------#
+
+from liota.core.package_manager import LiotaPackage
+import logging
+
+
+class PackageClass(LiotaPackage):
+
+    def __init__(self):
+        super(PackageClass, self).__init__()
```
KohliDev: The package class should use the device_comms rather than the device entity.

tompscanlan:
Thanks for the notes.  Closing as this was for discussion only.

PR #93, opened 2017-03-06
remove smartquotes

### LINKED ISSUE #92: smartquotes in README.md cause problems in setup.py install
I'm testing on various systems, and sometimes will get errors like this when running `python setup.py install`:

```bash
(venv) root [ /tmp/liota ]# python  setup.py  install
Traceback (most recent call last):
  File "setup.py", line 44, in <module>
    long_description = f.read()
  File "/tmp/liota/venv/lib/python3.5/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 4822: ordinal not in range(128)
```

Turns out the README has some smartquotes.  Removing those and replacing with simple quotes fixes the problem.

tompscanlan:
fixes #92 

Content ought to be the same, just special chars removed so various versions of python can read the README.md in setup.py without hitting errors


KohliDev:
Hello Tom, 

Thanks for the PR. Please resolve the branch conflicts and we`ll merge the pull request by tomorrow. 

Thanks,
Kohli

tompscanlan:
Conflicts ought to be gone

PR #94, opened 2017-03-06
add systemd units

tompscanlan:
add systemd start up configuration for liota


setup.py
```diff
@@ -125,8 +125,11 @@
                   'packages/examples/iotcc_thermistor_simulated.py']),
                 (os.path.abspath(os.sep) + '/../etc/liota/packages/edge_systems/dell5k',
                  ['packages/edge_systems/dell5k/edge_system.py']),
-		(os.path.abspath(os.sep) + '/../etc/liota/scripts',
+                (os.path.abspath(os.sep) + '/../etc/liota/scripts',
```
KohliDev: Please add the new dependencies in Manifest.in file

scripts/liotad.service
```diff
@@ -0,0 +1,11 @@
...
+# systemd unit configuration
+#
+[Unit]
+Description=Little Internet of Things Agent Daemon
+
+[Service]
+ExecStart=/etc/liota/packages/liotad.py
+
+[Install]
+WantedBy=multi-user.target
```
KohliDev: I`m not sure but I think better to add a new line in both the files liotad.service & liotad.socket

http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

tompscanlan:
Fixed endlines and added deps to the manifest file

README.md
```diff
@@ -110,6 +110,13 @@ To enable/disable the autostart service perform the following steps:
   $ chkconfig --del autostartliota
 ```
 
+### On systemd based systems:
+
+```bash
+  $ systemctl enable liota.socket
```
KohliDev: Should it be liotad.service & liotad.socket?

Also,  I cloned your branch and will try out this commands on Fedora VM.  What is the prime job of this commands? Also, I think if liota daemon gets started then we won`t require liotad.socket to be started explicitly.

KohliDev: @tompscanlan 

It'll be very helpful if you can provide some information about this change.

KohliDev:
@tompscanlan any update on this PR?

README.md
tompscanlan: You are correct.  Fixing that.

Issue #80, opened 2017-02-08
Detect correct network interfaces and "disks" in all the Liota examples 


KohliDev:
In all the liota examples related to health stats, pre-detect the network interfaces (eth0 or ens33) and disks(sda)

@Venkat2811

Venkat2811:
Addressed in #97

PR #97, opened 2017-03-08
Detect default network interface and disk name

Venkat2811:
Fix for #80


liota/lib/utilities/utility.py
```diff
@@ -110,6 +111,28 @@ def get_linux_version():
...
+    log.info("Default network interface is : {0}".format(nw_iface))
+    return nw_iface
+
+
+def get_disk_name():
+    """
+    Works with Linux.
+    :return: Disk type of the Edge_System
+    """
+    cmd = "lsblk -io KNAME,TYPE | grep 'disk' | sed -n '1p' | grep -o '^\S*'"
```
warthog9: What happens if the system has multiple disks installed on it?  I'm not using the system in question, but my laptop has two drives, and the one returned is sdb vs. sda.  Unsure if there's a good answer if there are multiple drives like this though. (just a note since I was reading the commit)

liota/lib/utilities/utility.py
```diff
@@ -110,6 +111,28 @@ def get_linux_version():
     return platform.platform()
 
 
+def get_default_network_interface():
+    """
+    Works with Linux.
+    :return: Default Network Interface of the Edge_System
+    """
+    cmd = "route | grep '^default' | grep -o '[^ ]*$'"
```
warthog9: There are situations where route may not actually return a default route in the main routing table, as the default route might be kept in another table.  This isn't a 'normal' circumstance, but something to be aware of

liota/lib/utilities/utility.py
Venkat2811: Thank you @warthog9 for pointing these out.  Tested these commands on DellGW 5000, RPi, Ubuntu, CentOS and RHEL VMs.  These utilities will be used when liota is collecting edge_system stats.  If an edge_system falls under these 'unusual' circumstance, I think it should be handled manually.

liota/lib/utilities/utility.py
```diff
@@ -1,4 +1,3 @@
-
 # -*- coding: utf-8 -*-
```
pmasrani: Raise an issue to split and rename this file. A collection of unrelated functions in "lib/utilities/utility" does not look good.

PR #98, opened 2017-03-12
Organization Group property changes

KohliDev:
The Organization Group property will be updated with help of Package manager.

Structured the naming mechanism in the package manager registry for reg_devices related to iotcc only.

### DIFF
--- a/liota/dccs/iotcc.py
+++ b/liota/dccs/iotcc.py
@@ -138,13 +138,13 @@ class IotControlCenter(DataCenterComponent):
                 raise RegistrationFailure()
             log.info("Resource Registered {0}".format(entity_obj.name))
             if entity_obj.entity_type == "HelixGateway":
-                self.store_reg_entity_details("EdgeSystem", entity_obj.name, self.reg_entity_id)
+                self.store_reg_entity_details(entity_obj.entity_type, entity_obj.name, self.reg_entity_id)
                 self.store_edge_system_uuid(entity_obj.name, self.reg_entity_id)
                 with self.file_ops_lock:
                     self.store_reg_entity_attributes("EdgeSystem", entity_obj.name,
                                                      self.reg_entity_id, None, None)
             else:
-                self.store_reg_entity_details("Devices", entity_obj.name, self.reg_entity_id)
+                self.store_reg_entity_details(entity_obj.entity_type, entity_obj.name, self.reg_entity_id)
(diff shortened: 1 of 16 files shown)


liota/dccs/iotcc.py
```diff
@@ -64,6 +64,7 @@ def __init__(self, username, password, con):
         self.username = username
         self.password = password
         self.proto = HelixProtocol(self.con, username, password)
+        self.info_file = self._init_info()
```
pmasrani: It is really difficult to understand what info is getting initialized here. Also, why on initialization a file is returned?

liota/dccs/iotcc.py
```diff
@@ -287,6 +305,57 @@ def prettify(self, elem):
         reparsed = minidom.parseString(rough_string)
         return reparsed.toprettyxml(indent="    ")
 
+    def _init_info(self):
+        msg = {
+            "iotcc": {
+                "EdgeSystem": {"SystemName": "", "EntityType": "", "uuid": ""},
+                "OGProperties": {"OrganizationGroup": ""},
```
pmasrani: Do we have multiple properties or only Organization Group Name? It can be confusing why we need OGProperties if it is just the Org Group name.

packages/examples/iotcc_resource_properties.py
```diff
@@ -0,0 +1,70 @@
...
+        # Acquire resources from registry
+        iotcc = registry.get("iotcc")
+
+        # Get values from configuration file
+        iotcc_json_path = read_liota_config('IOTCC_PATH', 'iotcc_path')
+        if iotcc_json_path == '':
+            return
+        try:
+            with open(iotcc_json_path, 'r') as f:
+                json_obj = json.load(f)["iotcc"]
```
KohliDev: Renamed to iotcc_details_json_obj

liota/dccs/iotcc.py
```diff
@@ -287,6 +302,56 @@ def prettify(self, elem):
         reparsed = minidom.parseString(rough_string)
         return reparsed.toprettyxml(indent="    ")
 
+    def _iotcc_json_structure(self):
```
pmasrani: It might be better to use "actions" as method names, like "_get_iotcc_json_structure"

KohliDev: Updated to _create_iotcc_json

PR #104, opened 2017-03-30
Proper Exception Hanlding and messages received from Queue

KohliDev:


liota/dccs/iotcc.py
```diff
@@ -109,29 +110,24 @@ def register(self, entity_obj):
...
-                                log.info("FOUND RESOURCE: {0}".format(json_msg["body"]["uuid"]))
-                                self.reg_entity_id = json_msg["body"]["uuid"]
-                            else:
-                                # TODO : Remove this code if required
-                                log.info("Waiting for resource creation")
-                                self.comms.send(json.dumps(
-                                    self._registration(self.next_id(), entity_obj.entity_id, entity_obj.name,
-                                                       entity_obj.entity_type)))
-                                on_response(self.recv_msg_queue.get())
+                    json_msg = json.loads(msg)
```
pmasrani: Can you please elaborate, what exactly was causing the problem and how it got resolved? I didn't understand it from the changes.

liota/dccs/iotcc.py
```diff
@@ -83,15 +83,16 @@ def on_response(msg):
                     return True
                 else:
                     log.debug("Processed msg: {0}".format(json_msg["type"]))
-                    on_response(self.recv_msg_queue.get())
+                    on_response(self.recv_msg_queue.get(True,10))
```
KohliDev: Yes if some other message comes in the queue

liota/dccs/iotcc.py
KohliDev: The problem was in retry logic loop which is resolved in this pull request.

liota/lib/transports/web_socket.py
```diff
@@ -56,6 +56,7 @@ def connect_soc(self):
             self.WebSocketConnection(self.url, False)
         except Exception:
             log.exception("WebSocket exception, please check the WebSocket address and try again.")
+            raise Exception("WebSocket exception, please check the WebSocket address and try again.")
```
KohliDev: Yes, I`ll remove log statements.

KohliDev: Over here exception might be of wide range and we need to catch all of them ranging from websocket object none to websocket close to websocket send exceptions. Thus, it is fine to catch all type of exceptions in typical scenarios. 

http://stackoverflow.com/questions/21938/is-it-really-that-bad-to-catch-a-general-exception

liota/dccs/iotcc.py
pmasrani: What other messages can we receive at the time of connection establishment?

KohliDev: create_relationship_Response

liota/lib/transports/web_socket.py
pmasrani: Catching "Exception" is fine, but while raising an exception, we should raise a specific one like LiotaException or LiotaConnectionException. Anyway, for this particular case, we can just remove the try catch block completely, as we are just rethrowing the caught exception.

What happens when the exception is thrown? Will liota program stop running? For package, i guess it will not get loaded (not affecting liotad), correct?

KohliDev: Liotad will never stop, the package won`t be loaded it is the general case and is tested in this scenario. And it's better to catch the exception and print the message around checking the Websocket address rather than just throwing a general exception.

liota/dccs/iotcc.py
pmasrani: How can we receive create_relationship_Response at time of establishment of a new connection? Is it during re-connection? Are you receiving it for both MQTT and Websocket? I believe it might be only for MQTT, as in MQTT we are re-using the topic which was used during previous connection. For Websocket, due to a different transaction id, a re-connection will not receive messages from previous connection.

KohliDev: It is a general logic in the lower usecase you`ll need retry for create_relationship over here to receive the other messages for HelixProtocol.

liota/lib/transports/web_socket.py
```diff
@@ -55,7 +55,7 @@ def connect_soc(self):
         try:
             self.WebSocketConnection(self.url, False)
         except Exception:
-            log.exception("WebSocket exception, please check the WebSocket address and try again.")
+            raise Exception("WebSocket exception, please check the WebSocket address and try again.")
```
pmasrani: I just noticed that we are losing the complete information about the actual exception, which can be an issue, if we are failing due to some other reason than what your message is suggesting. Here is why it is not a good idea: http://www.ianbicking.org/blog/2007/09/re-raising-exceptions.html

KohliDev: Traceback was getting printed as the exception was getting logged previously but you wanted it to be removed as per previous review comment. I have included           log.error(traceback.format_exc()) now it will satisfy both the use cases.

PR #105, opened 2017-04-04
Add sync with cloud adapter feature for Device Discovery:

winniex1:
1. for property related operations, if reboot occured (based on timestamp),
need to first get properties from cloud for loacl storage update.
2. only apply to iotcc dcc, and corresponding adapter needs to
support get properties request

### DIFF
--- a/liota/dccs/iotcc.py
+++ b/liota/dccs/iotcc.py
@@ -37,7 +37,9 @@ import threading
 import ConfigParser
 import os
 import Queue
+import datetime
 from time import gmtime, strftime
+from uptime import boottime
 from threading import Lock
 import xml.etree.cElementTree as ET
 from xml.dom import minidom
@@ -67,6 +69,7 @@ class IotControlCenter(DataCenterComponent):
         self._iotcc_json = self._create_iotcc_json()
         self.counter = 0
         self.recv_msg_queue = self.comms.userdata
+        self.boottime = boottime()
 
         self.dev_file_path = self._get_file_storage_path("dev_file_path")
         # Liota internal entity file system path special for iotcc
@@ -205,6 +208,13 @@ class IotControlCenter(DataCenterComponent):
         for key, value in properties.items():
             msg["body"]["property_data"].append({"propertyKey": key, "propertyValue": value})
         return msg
+
+    def _get_properties(self, msg_id, res_uuid):
+        return {
+            "transactionID": msg_id,
+            "type": "get_properties",
+            "uuid": res_uuid
+        }
 
     def _format_data(self, reg_metric):
         met_cnt = reg_metric.values.qsize()
@@ -358,7 +368,7 @@ class IotControlCenter(DataCenterComponent):
         if prop_dict is not None:
             for key in prop_dict.iterkeys():
                 value = prop_dict[key]
-                if key == 'entity type' or key == 'name' or key == 'device type':
+                if key == 'entity type' or key == 'name' or key == 'device type' or key == 'Entity_Timestamp':
                     continue
                 ET.SubElement(root, "attribute", name=key, value=value)
         # add time stamp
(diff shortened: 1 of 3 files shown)


liota/lib/protocols/helix_protocol.py
```diff
@@ -147,5 +147,11 @@ def on_receive(self, msg):
         if msg["type"] == "connection_response":
             require_field(msg["body"], "result")
             return True
+
+        elif msg["type"] == "get_properties_response":
```
KohliDev: I don`t think this message comes down as part of HelixProtocol, this file only handles HelixPrtocol messages so this code is not required over here.

winniex1: old version needs it (while debugging, found it).
just tested with new version, which does not need.
Delete it.

liota/dccs/iotcc.py
```diff
@@ -422,6 +432,7 @@ def store_device_info(self, uuid, name, dev_type, prop_dict):
 
     def write_entity_file(self, prop_dict, res_uuid):
         file_path = self.entity_file_path + '/' + res_uuid + '.json'
+        prop_dict.update({"Entity_Timestamp": datetime.datetime.now().strftime("%Y-%m-%dT%H:%M:%S")})
```
pmasrani: Do we have any specific reason why we are persisting time value in this format, instead of time since epoch? I believe even the timezone information is getting lost here, so can be an issue. Might be a good idea to use getUTCmillis() from utilities.

winniex1: Yes, the reason is that I need to get os restart time, but this is just for internal usage. The file with Entity_Timestamp is internal file, which will not be exposed to AW agents. The files AW agents need is under another folder, which will be deleted once AW agents get them.

uptime library is os independent, and do not have gcc dependency, so it is a very good choice. But its API only returns datetime. I need to compare current time with this datetime. So should not use getUTCmillis().

requirements.txt
```diff
@@ -3,4 +3,5 @@ linux-metrics==0.1.4
 pint==0.7.2
 paho-mqtt==1.2
 aenum==1.4.5
-CoAPthon==4.0.2
\ No newline at end of file
+CoAPthon==4.0.2
+uptime==3.0.1
```
KohliDev: hello @winniex1 

This is a new addition to Liota and will require approval from legal team. The process has already been completed please check with the Product team on how to get the approval for this new library.

liota/dccs/iotcc.py
pmasrani: uptime library comes with a big warning (Source:https://pythonhosted.org/uptime/):

"Warning On most platforms, this module depends very heavily on ctypes. It has become painfully apparent that many less mainstream platforms ship with a broken version of this standard library module, either accidentally or deliberately. Please test your Python installation before using uptime."

Have you tested it on gateways like DK300 and Dell 5000?

Regarding datetime, it can actually be converted to epoch using:

(datetime.datetime(2012,04,01,0,0) - datetime.datetime(1970,1,1)).total_seconds()
Source: http://stackoverflow.com/questions/11743019/convert-python-datetime-to-epoch-with-strftime

winniex1: Although it warns like that, it turned out that is not a problem for all the gateways we have.
I tried it on all the gateways in our lab (of course it works in my mac os, linux ubuntu VMs, etc.), and all are successfully installed.

requirements.txt
winniex1: I remembered when Christian collected this library requirement, Ravi's team member declared that in the future, we may need more libraries, and may ask again from legal team.
I'll send Christian an email to request that.

winniex1:
move to use a new pull request:
https://github.com/vmware/liota/pull/148
Property process improvement

PR #110, opened 2017-04-17
Create AW Device discovery file during unregister

pmasrani:


liota/dccs/iotcc.py
```diff
@@ -166,6 +166,8 @@ def on_response(msg):
         self.comms.send(json.dumps(self._unregistration(self.next_id(), entity_obj.reg_entity_id)))
         on_response(self.recv_msg_queue.get(True,20))
         self.remove_reg_entity_details(entity_obj.ref_entity.name, entity_obj.reg_entity_id)
+        if entity_obj.ref_entity.entity_type != "HelixGateway":
```
KohliDev: Hope there is no deletion flag for HelixGateway.

pmasrani: No, I discussed it with Arun. Gateway deletion from agent is not supported in Airwatch. Hence, there is no deletion flag for HelixGateway.

PR #112, opened 2017-04-20
Providing options for Cert Validation in IoTCC MQTT Packages

KohliDev:


examples/mqtt/dcc_comms/iotcc/sampleProp.conf
```diff
@@ -12,6 +12,12 @@ DevicePropList = {"Country":"USA-G", "State":"California", "City":"Palo Alto", "
...
 
 BrokerIP = "Broker Name or IP"
-BrokerPort = 1883
-broker_username = None
-broker_password = None
+BrokerPort = 8883
+broker_username = "Username"
+broker_password = "Password"
+cert_required = "CERT_REQUIRED"
+tls_version = "PROTOCOL_TLSv1"
```
pmasrani: I believe TLS v1 is deprecated and should not be used. Can't we use TLS 1.2?

KohliDev: I`ll check it, if we need to change I`ll address it in seperate PR

packages/sampleProp.conf
```diff
@@ -23,20 +23,20 @@ GraphitePort = None
 
 #### [GENERICMQTT] ####
 
-BrokerIP = "mqtt-broker"
-BrokerPort = 1883
-keep_alive = 60
-ConnectDisconnectTimeout = 10
+BrokerIP = "Broker Name or IP"
```
pmasrani: Minor nit, Hostname instead of name.

PR #120, opened 2017-05-04
Add Liota logging formatter to remove newlines

pmasrani:


config/logging.json
```diff
@@ -3,6 +3,7 @@
     "disable_existing_loggers": false,
     "formatters": {
         "simple": {
+            "()" : "liota.lib.utilities.log_formatter.LiotaLogFormatter",
```
KohliDev: Please verify the fix with the security team once.

liota/lib/utilities/log_formatter.py
```diff
@@ -0,0 +1,39 @@
...
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)    #
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF     #
+#  THE POSSIBILITY OF SUCH DAMAGE.                                            #
+# ----------------------------------------------------------------------------#
+import logging
+
+class LiotaLogFormatter(logging.Formatter):
+
+    def format(self, record):
+        record.msg = record.msg.replace('\n', '__\\n__').replace('\r', '__\\r__')
```
KohliDev: record.msg.strip() can be explored to be used.
https://docs.python.org/2/library/string.html

pmasrani: strip() removes leading and trailing characters only. Here, we need to remove the newline characters inserted somewhere in the middle of the message.

KohliDev: Strip also takes care of additional forgery including spaces, tabs, newlines and carriage returns. I believe it should be used. We should also check with the security team how to handle extra tabs if forged in between the log messages.

pmasrani: Can you point me to such documentation of strip? I didn't find it at: https://docs.python.org/2/library/string.html#string.strip.

Also, the security problem here is that, someone can introduce a complete new log message in liota, by putting arguments to the log messages as ...\nSOME_NEW_LOG_MESSAGE\n... and it will not be possible to figure out genuine vs these new logs. We can still check with the security team, nonetheless.

KohliDev: > $python
>  msg = "hello\t"
>  msg.strip()
> 'hello'

I think you can try it with code then referring the documentation.

Also, we need to check with the security if there are other characters then newline which might be used for forgery in log messages and required to be removed.

pmasrani: 1) The issue opened by security team mentions only newline character. Will check with security team if any other character can be a problem.
2) As I mentioned before, strip() takes care of only leading and trailing characters. Thus, it will not take care of the middle '\n' in the below example, which is the actual problem from security perspective:
 $python
msg = "\nhello\nworld\n"
msg.strip()
'hello\nworld'

KohliDev: Yes, it is known strip() will take care of leading and trailing characters only. You can apply "replace" operation "post" strip on a string if it is only about handling newline characters.

KohliDev:
As security team approved the changes merge it.

PR #121, opened 2017-05-04
Remove execfile usage from liota

pmasrani:


packages/sampleProp.conf
```diff
@@ -1,4 +1,4 @@
-#### [DEFAULT] ####
+[DEFAULT]
```
KohliDev: Do we require this to be removed from other sections as well ?, #### [IOTCC] ####
, #### [GENERICMQTT] ####

liota/lib/utilities/utility.py
```diff
@@ -272,6 +273,16 @@ def read_liota_config(section, name):
         log.warn('liota.conf file missing')
     return value
 
+def read_user_config(config_file_path):
+
+    config = ConfigParser.RawConfigParser()
+    config.optionxform = str
+    config.read(config_file_path)
+
+    user_config = dict(config.items('DEFAULT'))
```
KohliDev: Does it take care of getting the values from other sections in sampleprop.conf like 

IOTCC
GENERICMQTT

packages/sampleProp.conf
pmasrani: These are just comments in the conf file, as they begin with a '#' sign.

liota/lib/utilities/utility.py
pmasrani: We never made use of any sections in sampleProp.conf. IOTCC and GENERICMQTT
 are just comments, not configuration file sections. All the properties in sampleProp.conf belong only to DEFAULT section.

KohliDev: Okay, if we are keeping everything in "Default" section then the changes are correct.

KohliDev: If someone inserts new section tomorrow in sampleProp.conf, should we make user_config to hold data from all the sections? As this is utility method I believe we can keep it generic for our usage.

```
for each_section in conf.sections():
    for (each_key, each_val) in conf.items(each_section):
        print each_key
        print each_val
```

KohliDev: Otherwise, just add the comment that read_user_config method is used in our examples files only for reading keys in "Default" section.

pmasrani: As discussed, adding sections will add duplicate key issues. Will add the comment for read_user_config method about "Default" section.

PR #123, opened 2017-05-05
Certificate enablement in Websocket Comms and using Identity

KohliDev:


packages/sampleProp.conf
```diff
@@ -15,6 +15,10 @@ WebSocketUrl = "Websocket-address-url"
 IotCCUID = "Username"
 IotCCPassword = "Password"
 ShouldUnregisterOnUnload = "False"
+WebsocketCaCert = None
```
pmasrani: Is this a certificate path location? Probably renaming variable to indicate it is a path will help. 

Shouldn't we give example of the path? Otherwise it becomes very difficult to define it correctly, i.e., is it with quotes or without quotes etc.

In the longer term (not for this pull request), should we have a common variable like ServerCertificatePath for Websocket/MQTT/CoAP etc? The sampleProp file is getting cluttered with too many variables.

packages/sampleProp.conf
```diff
@@ -15,6 +15,10 @@ WebSocketUrl = "Websocket-address-url"
...
+VerifyCert= True
```
pmasrani: If we just change this to False, will the connection stop validating certificate?

packages/sampleProp.conf
```diff
@@ -15,6 +15,10 @@ WebSocketUrl = "Websocket-address-url"
...
+GatewayKey = None
+GatewayCert = None
```
pmasrani: This the client certificate path, correct? Please rename the variable.

examples/mqtt/device_comms/iotcc/samplePropMqtt.conf
```diff
@@ -15,6 +15,10 @@ DevicePropList = {"Country":"USA-G", "State":"California", "City":"Palo Alto", "
 WebSocketUrl = "Websocket-address-url"
 IotCCUID = "Username"
 IotCCPassword = "Password"
+WebsocketCaCert = None
+VerifyCert= False
```
pmasrani: I believe security team wanted certificate validation enabled by default. Please check with security team.

packages/sampleProp.conf
```diff
@@ -15,6 +15,10 @@ WebSocketUrl = "Websocket-address-url"
 IotCCUID = "Username"
 IotCCPassword = "Password"
 ShouldUnregisterOnUnload = "False"
+WebsocketCaCert = None
+VerifyCert= False
```
pmasrani: Should it be "verify server certificate"?

liota/lib/transports/web_socket.py
```diff
@@ -47,30 +47,39 @@ class WebSocket():
 
     """
 
-    def __init__(self, url):
+    def __init__(self, url,identity, verify_cert):
```
pmasrani: space missing afer url, in "url,identity"

examples/mqtt/device_comms/iotcc/samplePropMqtt.conf
```diff
@@ -15,6 +15,10 @@ DevicePropList = {"Country":"USA-G", "State":"California", "City":"Palo Alto", "
 WebSocketUrl = "Websocket-address-url"
```
pmasrani: Do we really need all these properties in samplePropMqtt.conf? We can have this file with only MQTT properties, as it is used only by MQTT examples, so that it is clear for the user.

packages/sampleProp.conf
KohliDev: Yes

examples/mqtt/device_comms/iotcc/samplePropMqtt.conf
KohliDev: Yes, because this example folder handles the connection to DeviceComms using MQTT and connection to DCCComms using Websocket.

examples/mqtt/device_comms/iotcc/samplePropMqtt.conf
KohliDev: Changed to True.

liota/lib/transports/web_socket.py
KohliDev: Fixed

liota/dcc_comms/websocket_dcc_comms.py
```diff
@@ -41,13 +41,15 @@
 
 class WebSocketDccComms(DCCComms):
 
-    def __init__(self, url):
+    def __init__(self, url, verify_cert, identity=None):
```
pmasrani: If verify_cert is True, identity should not be None. I think we should have this check somewhere.

liota/lib/transports/web_socket.py
```diff
@@ -46,32 +46,45 @@ class WebSocket():
...
+                else:
+                    log.error("Error : CA certificate path is missing")
+                    raise ValueError("Error : CA certificate path is missing")
+                if os.path.isfile(self.identity.root_ca_cert):
+                    try:
+                        self.ws = create_connection(self.url, enable_multithread=True,
+                                                    sslopt={"cert_reqs": ssl.CERT_REQUIRED,
+                                                            "ca_certs": self.identity.root_ca_cert})
+                    except ssl.SSLError:
+                        pass
```
pmasrani: Shouldn't we at least log this exception?

KohliDev: I have added the log statement. It's better if the review comments can be provided together on the entire codebase. Otherwise, they get missed and multiple updates have to be done. Post the first iteration only changes can be reviewed.

PR #126, opened 2017-05-09
With security consideration, improve Package Manager, Device Discovery, and IOT Control Center data storage

winniex1:
Package Manager Improvement:
1. restrict the command messenger of NamedPipe's access to only the owner for security.
2. do sanity check before load Liota Package: guarantee only packages under pkg_path in
liota.conf could be loaded.
3. verify Liota package integrity before execute it by passing down checksum to package manager.

Device Discovery Improvement:
1. for security consideration, now only MQTT and NamedPipe endpoint will be allowed.
a) restrict NamedPipe access to only the owner for security
b) since MQTT broker also sits on the edge system, MQTT subscriber can listen on
127.0.0.1/localhohst and use basic authentication (username and password) to guarantee
secured communication with MQTT broker. It's MQTT broker and MQTT publisher's responsibility
 to guarantee MQTT broker and external world communicate securely.
That is, by default, certification will be disabled and TLS will not be enforced.
2. restrict device discovery’s command messenger (NamedPipe) access to only the owner

IOT Control Center Improvement:
1. modify edge system attribute storage format: from xml file to json, because xml
libraries are not secure enough.


packages/dev_disc/README.md
```diff
@@ -64,9 +64,9 @@ devsim_cmd_msg_pipe = /var/tmp/liota/devsim_cmd_messenger.fifo	# the named pipe
 
 							# if no item in this list, Device Discovery will not be started
 
-disc_msg_pipe = /var/tmp/liota/discovery_messenger.fifo		# currently, only support these 4 types
+disc_msg_pipe = /var/tmp/liota/discovery_messenger.fifo		# currently, support these 4 types (currently,
```
secursive: Is the address correct? It seems that it was changed in other locations but not here.

packages/dev_disc/README.md
```diff
@@ -64,9 +64,9 @@ devsim_cmd_msg_pipe = /var/tmp/liota/devsim_cmd_messenger.fifo	# the named pipe
...
 
-socket = 127.0.0.1:5000						# you can only use some of them by deleting others
+socket = 127.0.0.1:5000						# coap and socket are not allowed for security consideration).
 
 mqtt = 127.0.0.1:1883:device_discovery				# IP address should be updated according to your system
```
secursive: If TLS/cert check is not used on loopback, the port number should be below 1024. Otherwise another process can connect to 1883, pretend to be MQTT endpoint and steal the credentials passed by clients.

liota/core/package_manager.py
```diff
@@ -1090,8 +1079,8 @@ def initialize():
                 log.error("Could not create directory for messenger pipe")
                 return
         try:
-            os.mkfifo(package_messenger_pipe)
-            log.info("Created pipe: " + package_messenger_pipe)
+            os.mkfifo(package_messenger_pipe, 6)
```
secursive: Incorrect permissions. Use 0600 as mode instead.

liota/lib/utilities/utility.py
```diff
@@ -313,7 +331,7 @@ def validate_named_pipe(self, pipe_file):
                     log.error("Could not create directory for messenger pipe")
                     return False
             try:
-                os.mkfifo(pipe_file)
+                os.mkfifo(pipe_file, 6)
```
secursive: Incorrect permissions. Use 0600 as mode instead.

packages/dev_disc/README.md
winniex1: updated

packages/dev_disc/README.md
winniex1: modified to 1023, and requested to use port < 1024

liota/lib/utilities/utility.py
winniex1: made a mistake; intended to restrict to the owner; corrected.

liota/core/package_manager.py
winniex1: corrected.

packages/dev_disc/README.md
```diff
@@ -58,33 +58,33 @@ Banana23 = iotcc	# Later user could create Liota Package for discovered devices
...
 							# if no item in this list, Device Discovery will not be started
 
-disc_msg_pipe = /var/tmp/liota/discovery_messenger.fifo		# currently, only support these 4 types
+disc_msg_pipe = /etc/liota/packages/dev_disc/discovery_messenger.fifo	# currently, support these 4 types (currently,
 
-socket = 127.0.0.1:5000						# you can only use some of them by deleting others
+socket = 127.0.0.1:5000						# coap and socket are not allowed for security consideration).
 
-mqtt = 127.0.0.1:1883:device_discovery				# IP address should be updated according to your system
+mqtt = 127.0.0.1:1023:device_discovery				# IP address should be updated according to your system
```
secursive: Why would IP address need to be updated for localhost? Instead the comment should suggest that if the IP address is changed to non-localhost IP, TLS should be enabled.

winniex1: that is old comments for socket and coap. Will delete that line for current codes.

winniex1: done

packages/examples/bike_simulator.py
```diff
@@ -38,9 +38,14 @@
...
 class PackageClass(LiotaPackage):
 
-    def run(self, registry):
+    def run(self, registry, package_record):
         from liota.entities.devices.bike_simulated import BikeSimulated
         import pint
+        from liota.lib.utilities.utility import check_integrity
+
+        # verify file integrity first
+        if (check_integrity(package_record, __file__) == False):
```
secursive: You are correct. Thanks for pointing it out.

This check should actually be done in package_manager near the beginning of _package_load method right after calculating the file hash. The original hash passed through the pipe should be stored and be compared with this hash. An alternative would be to place the check in _load_module_package.

winniex1: No, your mind is restricted to package manager. Let me provide you some examples of how sha1 checksum work for downloading a file by user and receiving a package from network interface.
1. The file provider will indicate SHa1 number in downloading operation, and when user finishes downloading, they need a tool to calculate sha1 checksum and compare it with that known checksum.
2. before TXer sends a package, then calculate checksum and insert it into the package; the receiver will extract that checksum, calculate checksum with the same way for the received package, and compare whether 2 checksums are the same.

Here for our package manager, it first gets loading cmd from user, and will do sanity check and figure out which file it will really load. Based on final filename, package manager can then calculate SHA1. 

Now, package manager will call the Liota Package's run() to execute, at that point, it should provide the SHA1 for the user to verify whether I am running the file that package manager is asking me to run.

winniex1: Reply to secursive:

No, that's not as you thought. When you pass in the loading cmd, you just provide the package name, NO file extensions. It is the Package manager who will do sanity check and decide which file exactly need to be loaded and fetched. Also please refer to my explanation/examples to Piyush.

winniex1: Sorry that after discussed with Greg, I agreed with you and will modify codes to let user pass in checksum.

config/liota.conf
```diff
@@ -17,11 +17,11 @@ collect_thread_pool_size = 30
 
 [PKG_CFG]
 pkg_path = /etc/liota/packages
-pkg_msg_pipe = /var/tmp/liota/package_messenger.fifo
+pkg_msg_pipe = /etc/liota/packages/package_messenger.fifo
```
winniex1: Currently, for package manager and device discovery related Named Pipe, they are restricted to be only accessible by the owner.

liota/core/package_manager.py
```diff
@@ -518,14 +515,23 @@ def _package_load(self, file_name, ext_forced=None, check_stack=None):
             log.warning("Package already loaded: %s" % file_name)
             return None
 
-        path_file_ext, file_ext = self._package_chk_exists(
+        path_file_ext, file_ext, checksum_list = self._package_chk_exists(
```
secursive: You can NOT calculate checksum twice from the file and just compare it. It will ALWAYS be the same.

For packages being loaded using 'load', 'update', 'reload' etc commands over pipe, you must receive the package hash along with file name and pass it to _package_load along-with filename.

For packages being automatically loaded using pkg_list file at startup, pkg_list file should contain hash values in it alongwith the filenames.

winniex1: My first thought was also like what you said. But we have not only those cmds, but also support pkg_list as its cmd input, and of course we also have packages_auto.txt which will have package list and will be loaded when package manager is started. So I had thought users might feel troublesome to have to calculate hash first and then fill them into that file. 

Anyway, I had tried to avoid above troublesome for both users and package manager.
But if you think that's the only way, I can then modify in that way.

liota/dccs/iotcc.py
```diff
@@ -390,44 +393,59 @@ def remove_reg_entity_details(self, entity_name, reg_entity_id):
...
+                json.dump(msg, f, sort_keys=True, indent=4, ensure_ascii=False)
+                log.debug('Initialized ' + file_path)
+            f.close()
+        except IOError, err:
+            log.error('Could not open {0} file '.format(file_path) + err)
 
-        log.debug("store_edge_system_info dev_file_path:{0}".format(self.dev_file_path))
-        file_path = self.dev_file_path + '/' + uuid + '.xml'
-        with open(file_path, "w") as fp:
-            fp.write(self.prettify(root))
```
secursive: If etree and minidom are not being used anymore, you can remove their import statements.

liota/core/device_discovery.py
```diff
@@ -471,36 +481,38 @@ def device_msg_process(self, data):
...
+                name = key + '_' + unique_key_value;
+                self._save_devinfo(name, key)
+                for dcc in key_dcc[:]:
+                    # register device to dcc, set properties for the device
+                    if 'iotcc' in dcc.lower():
+                        prop_dict = self.add_organization_properties("iotcc", value)
+                    else:
+                        prop_dict = value
+                    # TBM: temporarily assume dcc pkg will register edge_system as dcc name + "_edge_system"
+                    edge_system_pkg = dcc + str("_edge_system")
```
pmasrani: This assumption seems to be a bad idea. Please raise an issue to fix this in the long term.

packages/dev_disc/README.md
```diff
@@ -107,6 +105,8 @@ retry = 5
 keep_alive = 60
 
 ConnectDisconnectTimeout = 10
+* When MQTT broker also sits on the edge system, MQTT subscriber can listen on 127.0.0.1/localhost with port < 1024 and use basic authentication
```
pmasrani: Is it enforced that MQTT broker sits on the same edge system? Also, did you test by running the broker on port less than 1024, just to ensure no issues are faced?

liota/dccs/iotcc.py
pmasrani: Please remove prettify() function as well.

README.md
```diff
@@ -41,7 +41,7 @@ Liota supports plain old BSD sockets, WebSocket and MQTT communication protocols
 Liota applications can be broken into small pieces that can be loaded and unloaded into a running liota process. We recommend putting the EdgeSystems, Devices, Metrics and DCC(s) in separate packages. Then each construct can be loaded and unloaded at will. See the README in the package directory for complete details.
 
 ## Device Discovery
-Liota SDK provides users a way to discover new devices at run-time through a dedicated discovery thread. The discovery thread could listen on a list of end points by spinning up a listening thread for each of them. Currently, Liota supports 4 kinds of end points: MQTT, COAP, Socket, and Named Pipe.
+Liota SDK provides users a way to discover new devices at run-time through a dedicated discovery thread. The discovery thread could listen on a list of end points by spinning up a listening thread for each of them. Currently, Liota supports 4 kinds of end points: MQTT, COAP, Socket, and Named Pipe. However, because of security consideration on edge system, currently only MQTT and Named Pipe are allowed, especially Named Pipe is only accessible to its owner user.
```
pmasrani: Do we have plans to support COAP and sockets soon, else it might be good to remove from documentation and config files for now, and introduce it later when we actually introduce the support.

config/liota.conf
winniex1: Got it.

liota/core/package_manager.py
winniex1: done

liota/dccs/iotcc.py
winniex1: done

liota/core/device_discovery.py
winniex1: Agree. that's also why I make a mark as TBM. Just currently have no better way.

README.md
winniex1: CoAP will be used when its library supports DTLS;
Socket could be used by internal test.
Therefore, it's better to keep codes there while disable them for pip liota install.

liota/core/package_manager.py
```diff
@@ -470,7 +495,7 @@ def _package_chk_exists(self, file_name, ext_forced=None):
             return None, None
```
secursive: Should be None, None, None or None, None, [].

liota/core/package_manager.py
```diff
@@ -448,9 +461,21 @@ def _package_chk_exists(self, file_name, ext_forced=None):
...
+            log.error("Package %s is NOT under package path %s"
+                         % (file_name, package_path))
+            return None, None
 
         file_ext = None
         extensions = ["py", "pyc", "pyo"]
+
+        # get checksum for all candidates: path_file + ".py/.pyc/.pyo"
+        checksum_list = []
+        for ext in extensions:
```
secursive: It would probably be better to only return the checksum of path_file_ext that is returned at the end of function and not of all extensions. However, this needs to be thought about and can be done in a later pull request.

packages/dev_disc/README.md
pmasrani: Any update on this?

liota/core/package_manager.py
```diff
@@ -376,37 +359,67 @@ def run(self):
...
                 #-----------------------------------------------------------
                 # Use these commands to handle package management tasks
 
                 with package_lock:
                     if len(msg) < 2:
                         log.warning("No package is specified: %s" % command)
                         continue
-                    if len(msg) > 2:
-                        list_packages = msg[1:]
+                    if len(msg) > 3:
```
pmasrani: What if length of message is equal to 2? I see a check for length < 2 and then directly for greater than 3, but no check for length = 2?

liota/core/package_manager.py
```diff
@@ -376,37 +359,67 @@ def run(self):
...
+                        list_arg = msg[1:]
+                        list_packages = []
+                        cnt = len(list_arg)
+                        i = 0
+                        while (i < cnt):
+                            list_packages.append({list_arg[i]: list_arg[i+1]})
```
pmasrani: What if the number of arguments is odd? Say, somebody forgot to mention checksum for the last package name?

liota/core/package_manager.py
```diff
@@ -509,23 +534,33 @@ def _package_module_load(self, file_name, path_file_ext, file_ext):
     # This method is called to load package into current Liota process using
     # file_name (no_ext) as package identifier.
 
-    def _package_load(self, file_name, ext_forced=None, check_stack=None):
+    def _package_load(self, file_name, checksum=None, ext_forced=None, check_stack=None):
```
pmasrani: I think checksum should be a mandatory argument, not optional with default as None.

I believe it is optional right now because while loading dependencies, this method is getting called, without checksum parameter. I believe this can also be a security issue. What if someone changes a dependency package and loads a trusted package which loads its modified dependency file? Please check with security team.

packages/README.md
```diff
@@ -9,9 +9,9 @@ Currently, supported commands include package action commands and statistical co
 
 ###Package action commands
 
-* **load** package_name [package_name] ...
+* **load** package_name sha1_checksum [package_name sha1_checksum] ...
```
pmasrani: Should we also document or provide a utility to the end-user to calculate checksum. Thus, if he writes a package, he can run the utility to calculate checksum using our utility method and then use that checksum to load the package.

liota/core/package_manager.py
winniex1: done

liota/core/package_manager.py
winniex1: The dependency package should be loaded but we cannot expect users to list all those dependency packages' checksums. One way I thought out is to simply refuse that kind of loading request to force users to load packages in order. However, we will lose our flexibility. Therefor, I choose not to decline. If this will be new issue, will use another pull request to fix it. Meanwhile, will talk with security team.

packages/README.md
winniex1: For linux os, it is quite simply to get a checksum, just using "sha1sum filename". I pointed this out in commit message. Will write this in README.md.

liota/core/package_manager.py
winniex1: Checksum check will fail, and that package will not be successfully loaded. It may affect other packages' loading since it may mess up the other package name and checksum. It will not be loaded until the correct inputs are given.

liota/core/device_discovery.py
```diff
@@ -471,36 +481,38 @@ def device_msg_process(self, data):
...
+                    (dev, reg_dev) = self.reg_device(dcc, edge_system_pkg, name, key, prop_dict)
+                    if (dev is not None) and (reg_dev is not None):
+                        reg_rec = {}
+                        # add it in Records: currently only for debugging
+                        reg_rec[dcc] = (dev, reg_dev)
+                        self._update_devinfo(name, reg_rec)
+        except:
+            log.exception("device_msg_process error:{0}".format(sys.exc_info()[0]))
```
winniex1: I re-thought it over, and I have to change it back to log.error(""). An exception when process a received device message is expected (it may occur), and we just need to report it, and device discovery thread can still listen and process more received messages.

liota/core/device_discovery.py
winniex1: done

liota/core/package_manager.py
winniex1: I had ever thought of this issue.
path_file_ext could be .py /.pyc or .pyo, which is decided by package manager. The user only specify the Package Name without extension. Therefor, we need to use checksum_list.

liota/core/package_manager.py
winniex1: I checked ==2 case, after >3, have you seen?
                    if len(msg) == 2:
                        log.warning("No checksum of {0} is specified: {1}".
                                    format(file_name, command))

packages/dev_disc/README.md
winniex1: security team asked me to use port < 1024 to be securer. However, I tested many times on my mac. If we set listen port < 1024 for MQTT broker, the connection from any client (a test tool) will be rejected.
Of course, our liota dev_comms connection will also be rejected.
I'll modify that back, while mention in README.md that users had better set a special port # rather than well-known 1883.

winniex1:
Because unknown reason, new modifications could not be merged to this pull request.
Have to merge this pull request first to create next pull request to add another modification commit.

liota/core/package_manager.py
pmasrani: I don't think it will even reach checksum calculations. In case of odd number of arguments, accessing list_arg[i+1] will give an index out of range error. I think we should add a check to check the number of arguments before accessing (i+1)th element of the list.

packages/dev_disc/README.md
pmasrani: Please validate with security team if they are fine with any other port > 1024.

PR #131, opened 2017-05-17
Package Manager and Device Discovery Improvement (6):

winniex1:
1. When mqtt broker listens on a port < 1024, client connection will
be refused, so remind user to listen on a unique port rather than
well-known port 1883
2. correct _package_chk_exists() return None minor issue
3. improve exception logging


winniex1:
This is part of commits in pull request of #126, which has been approved by security team to let them work on verification.
If more issues are found later, will create new pull request.

PR #146, opened 2017-08-27
change path of liota generated files; remove bash command requirement

winniex1:
1. Replace Bash with Bourne shell syntax to remove bash command dependancy
2. change path of liota generated temporary files to the "/var/tmp" directory,
and reusable files to the "/usr/lib/" directory:
	a) "/etc/liota/conf/devs" to "/var/tmp/liota/devs"
	b) "/etc/liota/conf/entity" to "/usr/lib/liota/entity"
	c) "/etc/liota/conf/iotcc.json" to "/var/tmp/liota/iotcc.json"
3. remove "uuid.ini" since "iotcc.json" stores edge system and devices info

### DIFF
--- a/config/liota.conf
+++ b/config/liota.conf
@@ -4,13 +4,10 @@ json_path = /etc/liota/logging.json
 [LOG_PATH]
 log_path = /var/log/liota
 
-[UUID_PATH]
-uuid_path = /etc/liota/uuid.ini
-
 [IOTCC_PATH]
-dev_file_path = /etc/liota/conf/devs
-entity_file_path = /etc/liota/conf/entity
-iotcc_path = /etc/liota/conf/iotcc.json
+dev_file_path = /var/tmp/liota/devs
+entity_file_path = /usr/lib/liota/entity
+iotcc_path = /var/tmp/liota/iotcc.json
 
 [CORE_CFG]
 collect_thread_pool_size = 30
@@ -21,7 +18,7 @@ pkg_msg_pipe = /var/tmp/liota/package_messenger.fifo
 pkg_list = /usr/lib/liota/packages/packages_auto.txt
 
 [DISC_CFG]
-disc_cmd_msg_pipe = /usr/lib/liota/packages/dev_disc/disc_cmd_messenger.fifo
+disc_cmd_msg_pipe = /var/tmp/liota/packages/dev_disc/disc_cmd_messenger.fifo
 
 [DEVICE_TYPE_TO_UNIQUEKEY_MAPPING]
 Press64 = serial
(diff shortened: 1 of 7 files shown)


config/liota.conf
```diff
@@ -4,13 +4,10 @@ json_path = /etc/liota/logging.json
...
 log_path = /var/log/liota
 
-[UUID_PATH]
-uuid_path = /etc/liota/uuid.ini
-
 [IOTCC_PATH]
-dev_file_path = /etc/liota/conf/devs
-entity_file_path = /etc/liota/conf/entity
-iotcc_path = /etc/liota/conf/iotcc.json
+dev_file_path = /var/tmp/liota/devs
```
KohliDev: Let`s try to keep one path either "/usr/lib/liota" or "/var/tmp/liota". Different directories for files emitted by Liota will be difficult to manage at un-installation.

winniex1: use /usr/lib/liota

config/liota.conf
```diff
@@ -36,10 +33,10 @@ Apple56 = iotcc_mqtt, iotcc
 Banana23 = iotcc
 
 [DEVSIM_CFG]
-devsim_cmd_msg_pipe = /usr/lib/liota/packages/dev_disc/devsim_cmd_messenger.fifo
+devsim_cmd_msg_pipe = /var/tmp/liota/packages/dev_disc/devsim_cmd_messenger.fifo
```
KohliDev: You mentioned to me before the security team needed them to be in the same location as packages for better secuity. So, I don`t think we should remove them from "/usr/lib/liota" as packages and example now exist in this directory.

packages/liotapkg.sh
```diff
@@ -34,30 +34,32 @@
...
+if [ ! -f "$liota_config" ]
+then
     echo "ERROR: Configuration file not found" >&2
     echo "You made need to copy the distributed configuration file from /usr/lib/liota/config/liota.conf to /etc/liota/liota.conf" >&2
     exit -1
 fi
 
 while read line # Read configurations from file
 do
-    if echo $line | grep -F = &>/dev/null
```
KohliDev: Can this check be removed ?

winniex1: It can be removed.
I had tested.
This paragraph is just to get pkg_msg_pipe value. Without it, we still can make it well.

config/liota.conf
winniex1: I was told by you that OSS team told you to put liota generated temporary files under /var/tmp folder.
Although security team suggested to put them in the same location, you worked with OSS team and told me above.
I also saw someone (I guess you) modified pkg_msg_pipe back to /var/tmp/liota/package_messenger.fifo
So I just followed the way you did for pkg_msg_pipe. The pipe for device discovery and device simulator should follow the same procedure.
===
[PKG_CFG]
pkg_path = /usr/lib/liota/packages
pkg_msg_pipe = /var/tmp/liota/package_messenger.fifo
pkg_list = /usr/lib/liota/packages/packages_auto.txt

KohliDev: The temporary file I previously mentioned were devs,entity and iotcc.json which should exist in "/var/tmp/liota" folder. I never meant the pipe files as you already mentioned before that security team wanted it to be in the same location as the packages or example files. It was modified by John to "/var/tmp/liota/package_messenger.fifo" as there was the disparity in the README which caused him confusion. I think security team will not approve the pipe files in "/var/tmp/liota", if they have suggested they should exist in the same path as packages and examples files.

winniex1: ok. if it is because of misleading communication, will put pipes under /usr/lib

PR #147, opened 2017-09-01
improvement iotcc.json process

winniex1:
For IOTCC, add retry loading logic for iotcc.json file; if loading still fails, print out to let user be aware; for large scale device addition, make continuous addition possible when some device registration fails.

### DIFF
--- a/liota/dccs/iotcc.py
+++ b/liota/dccs/iotcc.py
@@ -169,11 +169,11 @@ class IotControlCenter(DataCenterComponent):
 
         self.comms.send(json.dumps(self._unregistration(self.next_id(), entity_obj.ref_entity)))
         on_response(self.recv_msg_queue.get(True, 20))
-        self.remove_reg_entity_details(entity_obj.ref_entity.name, entity_obj.reg_entity_id)
         if entity_obj.ref_entity.entity_type != "HelixGateway":
             self.store_device_info(entity_obj.reg_entity_id, entity_obj.ref_entity.name,
                                    entity_obj.ref_entity.entity_type, None, True)
         else:
+            self.remove_reg_entity_details(entity_obj.ref_entity.name, entity_obj.reg_entity_id)
             self.store_device_info(entity_obj.reg_entity_id, entity_obj.ref_entity.name, None, None, True)
 
         log.info("Unregistration of resource {0} with IoTCC complete".format(entity_obj.ref_entity.name))
--- a/liota/dccs/iotcc.py
+++ b/liota/dccs/iotcc.py
@@ -73,6 +73,7 @@ class IotControlCenter(DataCenterComponent):
         time.sleep(0.5)
         self.proto = HelixProtocol(self.comms, self.comms.identity.username, self.comms.identity.password)
         self._iotcc_json = self._create_iotcc_json()
+        self._iotcc_json_load_retry = int(read_liota_config('IOTCC', 'iotcc_load_retry'))
         self.counter = 0
         self.recv_msg_queue = self.comms.userdata
         self.dev_file_path = self._get_file_storage_path("dev_file_path")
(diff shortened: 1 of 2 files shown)


liota/dccs/iotcc.py
```diff
@@ -170,11 +169,11 @@ def on_response(msg):
 
         self.comms.send(json.dumps(self._unregistration(self.next_id(), entity_obj.ref_entity)))
         on_response(self.recv_msg_queue.get(True, 20))
-        self.remove_reg_entity_details(entity_obj.ref_entity.name, entity_obj.reg_entity_id)
         if entity_obj.ref_entity.entity_type != "HelixGateway":
             self.store_device_info(entity_obj.reg_entity_id, entity_obj.ref_entity.name,
                                    entity_obj.ref_entity.entity_type, None, True)
         else:
+            self.remove_reg_entity_details(entity_obj.ref_entity.name, entity_obj.reg_entity_id)
```
KohliDev: Can you please mention the reason to move this call in else loop for my understanding?

winniex1: now we only save edge system info in iotcc.json, no device info any more.
so should not do it devices. (!= "HelixGateway" is for devices).
If do, it will be in vain, and print some "cannot find xxx device info" warning/error, which is not what we want.

KohliDev: Okay and but do we support any entity with device_type as none, I think we can get rid of else loop?

liota/dccs/iotcc.py
```diff
@@ -332,16 +331,33 @@ def _create_iotcc_json(self):
...
     def store_reg_entity_details(self, entity_type, entity_name, reg_entity_id, entity_local_uuid):
-        msg = ''
         if self._iotcc_json == '':
             log.warn('iotcc.json file missing')
             return
         try:
-            with open(self._iotcc_json, 'r') as f:
-                msg = json.load(f)
-            f.close()
+            f = open(self._iotcc_json, 'r')
```
KohliDev: I believe you were going to add the lock mechanism over here.

liota/dccs/iotcc.py
```diff
@@ -332,16 +331,33 @@ def _create_iotcc_json(self):
...
         except IOError, err:
-            log.error('Could not open {0} file '.format(self._iotcc_json) + str(err))
+            log.exception('Could not open {0} file '.format(self._iotcc_json) + str(err))
+            return
+
+        def load_json_record(f):
```
KohliDev: An alternative mechanism to read iotcc_json file should be explored if the issue doesn`t get fixed with the retry mechanism.

liota/dccs/iotcc.py
winniex1: No, we need to generate uuid.json file with "remove": true, so need that if...else to do that.

liota/dccs/iotcc.py
winniex1: No, I uses a lock before calling this func rather than inside it. 
        if entity_obj.entity_type == "HelixGateway":
 -                self.store_reg_entity_details(entity_obj.entity_type, entity_obj.name, self.reg_entity_id,
 -                                              entity_obj.entity_id)
                  with self.file_ops_lock:
 +                    self.store_reg_entity_details(entity_obj.entity_type, entity_obj.name, self.reg_entity_id,
 +                                              entity_obj.entity_id)
                      self.store_reg_entity_attributes("EdgeSystem", entity_obj.name,
                                                       self.reg_entity_id, None, None)

liota/dccs/iotcc.py
winniex1: from my test, resolved. we only call this func when register edge system. but will see QE result.

liota/dccs/iotcc.py
KohliDev: I get the point but do we also store the information if the Helixgateway is deleted as the else is currently handling that.

liota/dccs/iotcc.py
KohliDev: Okay

liota/dccs/iotcc.py
winniex1: I think we still need it when we only delete some devices with helix gateway is still alive (still need to call store_device_info with specific parameters.)

liota/dccs/iotcc.py
```diff
@@ -73,6 +73,7 @@ def __init__(self, con):
         time.sleep(0.5)
         self.proto = HelixProtocol(self.comms, self.comms.identity.username, self.comms.identity.password)
         self._iotcc_json = self._create_iotcc_json()
+        self._iotcc_json_load_retry = int(read_liota_config('IOTCC', 'iotcc_load_retry'))
```
KohliDev: In my testing, I figured out I`m getting the following error after loading the package "iotcc_mqtt.py"
 Exception in initialization: No section: 'IOTCC'. Can you please look into the issue as I need to merrge my PR on top of this.

winniex1:
The bad thing is that my laptop does not work right now.
I am in IT helper to loan a computer.
If you can fix and submit a pull request, that’ll be very helpful.

Thx,
Winnie

KohliDev:
I`ll look into the issue as possible but it is better if you can go through it once and fix the issue with testing.

Thanks,
Kohli

PR #148, opened 2017-09-04
Property process improvement:

winniex1:
1. add get property feature support for iota dcc to sync local records
		with those in adapter when reboot occurs
    2. make the feature of getting property after reboot configurable (off by default)
    3. filter out system generated properties after calling "get properties"
    4. add process for a case when device type == ''


liota/dccs/iotcc.py
```diff
@@ -513,8 +540,16 @@ def store_reg_entity_attributes(self, entity_type, entity_name, reg_entity_id,
                 new_prop_dict = tmp_dict
         else:
             tmp_dict = self.read_entity_file(reg_entity_id)
-            if ((tmp_dict["entity type"] == entity_type) and (tmp_dict["name"] == entity_name)
-                and (tmp_dict["device type"] == dev_type)):
+            # check Entity_Timestamp of entity_file: if < boottime, get properties from cloud
+            if (self.enable_reboot_getprop == "True") and ('Entity_Timestamp' in tmp_dict):
+                last_dtime = datetime.datetime.strptime(tmp_dict["Entity_Timestamp"], "%Y-%m-%dT%H:%M:%S")
+                if (last_dtime <= self.boottime):
+                    list_prop = self.get_properties(reg_entity_id)
```
pmasrani: If i understand correctly, store_reg_entity_attributes() gets called for every set_properties() call. So if the gateway reboots and a user package has made 20 set_properties() call, your code will call get_properties() every time set_properties() call is made? Isn't that too many calls for the same entity?

What if user had set properties using a package, which is not to be loaded after reboot? Get_properties call will never be made?

liota/dccs/iotcc.py
```diff
@@ -590,3 +625,28 @@ def _unregistration(self, msg_id, ref_entity):
...
+                log.debug("Received msg: {0}".format(msg))
+                json_msg = json.loads(msg)
+                log.debug("Processed msg: {0}".format(json_msg["type"]))
+                if json_msg["type"] == "get_properties_response" and json_msg["body"]["uuid"] != "null" and \
+                                json_msg["body"]["uuid"] == resource_uuid:
+                    log.info("FOUND PROPERTIE LIST: {0}".format(json_msg["body"]["propertyList"]))
+                    self.prop_list = json_msg["body"]["propertyList"]
+                    log.info("prop_list:{0}".format(self.prop_list))
+                else:
+                    log.info("Waiting for getting properties")
```
pmasrani: Print the message which is getting ignored.

liota/dccs/iotcc.py
```diff
@@ -590,3 +625,28 @@ def _unregistration(self, msg_id, ref_entity):
...
+                    on_response(self.recv_msg_queue.get(True,30))
```
pmasrani: We are using wit time of 300 seconds at other places. Why is it 30 seconds here?

liota/dccs/iotcc.py
```diff
@@ -590,3 +625,28 @@ def _unregistration(self, msg_id, ref_entity):
...
+            except:
+                raise Exception("Exception while getting properties")
+
+        self.comms.send(json.dumps(self._get_properties(self.next_id(), resource_uuid)))
+        on_response(self.recv_msg_queue.get(True,30))
```
pmasrani: ditto

liota/dccs/iotcc.py
```diff
@@ -590,3 +625,28 @@ def _unregistration(self, msg_id, ref_entity):
...
+        self.prop_list = None
+
+        def on_response(msg):
+            try:
+                log.debug("Received msg: {0}".format(msg))
+                json_msg = json.loads(msg)
+                log.debug("Processed msg: {0}".format(json_msg["type"]))
+                if json_msg["type"] == "get_properties_response" and json_msg["body"]["uuid"] != "null" and \
+                                json_msg["body"]["uuid"] == resource_uuid:
+                    log.info("FOUND PROPERTIE LIST: {0}".format(json_msg["body"]["propertyList"]))
```
pmasrani: Incorrect spelling: PROPERTIE

liota/dccs/iotcc.py
```diff
@@ -590,3 +625,28 @@ def _unregistration(self, msg_id, ref_entity):
...
+                    self.prop_list = json_msg["body"]["propertyList"]
+                    log.info("prop_list:{0}".format(self.prop_list))
+                else:
+                    log.info("Waiting for getting properties")
+                    on_response(self.recv_msg_queue.get(True,30))
+            except:
+                raise Exception("Exception while getting properties")
```
pmasrani: Does this print the stacktrace of the original exception?

liota/dccs/iotcc.py
```diff
@@ -513,8 +540,16 @@ def store_reg_entity_attributes(self, entity_type, entity_name, reg_entity_id,
...
             tmp_dict = self.read_entity_file(reg_entity_id)
-            if ((tmp_dict["entity type"] == entity_type) and (tmp_dict["name"] == entity_name)
-                and (tmp_dict["device type"] == dev_type)):
+            # check Entity_Timestamp of entity_file: if < boottime, get properties from cloud
+            if (self.enable_reboot_getprop == "True") and ('Entity_Timestamp' in tmp_dict):
+                last_dtime = datetime.datetime.strptime(tmp_dict["Entity_Timestamp"], "%Y-%m-%dT%H:%M:%S")
+                if (last_dtime <= self.boottime):
+                    list_prop = self.get_properties(reg_entity_id)
+                    tmp_dict = self.merge_prop_dict_list(tmp_dict, list_prop)
+            if ((('entity type' in tmp_dict) and (tmp_dict["entity type"] == entity_type)) and
```
pmasrani: Can you please explain this condition in comments in the code? It is very hard to understand.

liota/dccs/iotcc.py
```diff
@@ -488,10 +501,24 @@ def read_entity_file(self, res_uuid):
...
+    def merge_prop_dict_list(self, prop_dict, prop_list):
+        # prop_dict: new property dictionary
+        # prop_list: list of dictionary items
+        if (prop_list is None):
+            return prop_dict
+        if (prop_dict is None):
+            prop_dict = {}
+        for item in prop_list:
+            for key in item:
+                if key.find("System Properties|") == -1:
```
pmasrani: Why are we skipping System Properties? Can we assume they are already in sync with Airwatch?

liota/dccs/iotcc.py
```diff
@@ -400,7 +412,7 @@ def write_entity_json_file(self, prop_dict, attribute_list, uuid, remove):
         if prop_dict is not None:
             for key in prop_dict.iterkeys():
                 value = prop_dict[key]
-                if key == 'entity type' or key == 'name' or key == 'device type':
+                if key == 'entity type' or key == 'name' or key == 'device type' or key == 'Entity_Timestamp':
```
pmasrani: what is the reason for this change?

liota/dccs/iotcc.py
winniex1: done

liota/dccs/iotcc.py
winniex1: done

liota/dccs/iotcc.py
winniex1: done

liota/dccs/iotcc.py
winniex1: changed to log.exception, and return None.

liota/dccs/iotcc.py
winniex1: This fun writes uuid.json file under devs folder (configurable) for AW to take and remove. It is based on local record bsed on entity folder (configurable). In this pull request, I added Entity_Timestamp to assist judgement of restarting, which is needed in local records. But it is not needed for uuid.json file for AW. Therefore, has to skip it.

liota/dccs/iotcc.py
winniex1: This "System Properties|xxx" are generated by vRops itself. If you get_properties(), you can get them. In fact, you can even see it through vRops UI.  But these are definitely not what we and AW want. So I filter them out.

liota/dccs/iotcc.py
winniex1: store_reg_entity_attributes() will be called not only when you call set_properties(), but when you call register().
In liota package, we always have to call register() to create or find the resources in HelixAdapter. Therefore, after reboot, get_properties() can at least be called once.

In fact, it will only be called ONCE, even you have 20 set_properties(). The reason is that whenever I modify the local record under entity folder (configurable), I will put the current time to "Entity_Timestamp". 
Then, Line 543-546 can guarantee get_properties will not be called any more, unless you have another reboot.

liota/dccs/iotcc.py
winniex1: done

liota/dccs/iotcc.py
pmasrani: Liota package is a user code. We cannot force a user to always call register() or set_properties(). You can not assume that get_properties() will be called.

liota/dccs/iotcc.py
```diff
@@ -536,8 +563,17 @@ def store_reg_entity_attributes(self, entity_type, entity_name, reg_entity_id,
...
-            if ((tmp_dict["entity type"] == entity_type) and (tmp_dict["name"] == entity_name)
-                and (tmp_dict["device type"] == dev_type)):
+            # check Entity_Timestamp of entity_file: if < boottime, get properties from cloud
+            if (self.enable_reboot_getprop == "True") and ('Entity_Timestamp' in tmp_dict):
+                last_dtime = datetime.datetime.strptime(tmp_dict["Entity_Timestamp"], "%Y-%m-%dT%H:%M:%S")
+                if (last_dtime <= self.boottime):
+                    list_prop = self.get_properties(reg_entity_id)
+                    # merge property info from get_properties() into our local entity record
+                    tmp_dict = self.merge_prop_dict_list(tmp_dict, list_prop)
+            if ((('entity type' in tmp_dict) and (tmp_dict["entity type"] == entity_type)) and
```
pmasrani: Can you please explain this if condition? It is so hard to understand. We are checking for None for some, '' for some, there are a lot of and and or statements...

liota/dccs/iotcc.py
```diff
@@ -240,6 +245,13 @@ def _properties(self, msg_id, entity_type, entity_id, entity_name, timestamp, pr
             msg["body"]["property_data"].append({"propertyKey": key, "propertyValue": value})
         return msg
 
+    def _get_properties(self, msg_id, res_uuid):
+        return {
+            "transactionID": msg_id,
+            "type": "get_properties",
```
pmasrani: Can you please change the type to get_properties_resquest to match to get_properties_response?

requirements.txt
```diff
@@ -5,4 +5,4 @@ mock==2.0.0
 paho-mqtt==1.2
 pint==0.7.2
 websocket-client==0.37.0
-
+uptime==3.0.1
```
pmasrani: There is one more requirements.txt for liota-lite. You will need to modify that too.

liota/dccs/iotcc.py
winniex1: If no register() nor set_properties() are called, AW will never get a uuid.json file to take. Nothing will be impacted. I did not assume user to call register() or set_properties(). But those are the only chance AW agent can get a new uuid.json file to take. When that happens, if reboot had occurred,  get_properties() will be called to make AW agent to have a chance to have updated/synced record of properties from voprs.

liota/dccs/iotcc.py
winniex1: yes

requirements.txt
winniex1: yes

liota/dccs/iotcc.py
winniex1: In fact, it is quite straightforward. It checked 'entity type', 'entity_name', 'device type' of this fun's inputs with those inside local record, and see whether they are for the same entity or not.
(('entity type' in tmp_dict) and (tmp_dict["entity type"] == entity_type)) is for 'entity type'.
front half is guarantee that 'entity type' is inside the local record.

For the 'entity name', it is similar.
For 'device type', it is a little different since we have "EdgeSystem" and "Connected Device". For 'EdgeSystem', our local record will be '' and the input dev_type should be None.

In fact, I have the comment under the if () # the same entity.

liota/dccs/iotcc.py
winniex1: moved it into helix adapter.

winniex1:
Yes, I’ll pull, merge do the final test again.

Winnie

winniex1:
have made the final test, and it works.

PR #163, opened 2017-10-17
Certificate Revocation List Check in Liota

KohliDev:
CRLs(Certificate Revocation Lists) check has been implemented:
https://docs.python.org/2/library/ssl.html#ssl.VERIFY_CRL_CHECK_CHAIN
 in this mode CRLs of all certificates in the peer cert chains are checked.
In order to enable the CRLs check pass the parameter 'crl_path' during TLSConf initialization, it requires a valid path to CRL in PEM format.

### DIFF
--- a/config/liota.conf
+++ b/config/liota.conf
@@ -10,6 +10,9 @@ entity_file_path = /usr/lib/liota/entity
 iotcc_path = /usr/lib/liota/iotcc.json
 iotcc_load_retry = 3
 
+[CRL_PATH]
+crl_path = None
+
 [CORE_CFG]
 collect_thread_pool_size = 30
 
--- a/config/liota.conf
+++ b/config/liota.conf
@@ -11,7 +11,7 @@ iotcc_path = /usr/lib/liota/iotcc.json
 iotcc_load_retry = 3
 
 [CRL_PATH]
-crl_path = None
+crl_path =
 
 [CORE_CFG]
 collect_thread_pool_size = 30
--- a/liota/lib/transports/mqtt.py
+++ b/liota/lib/transports/mqtt.py
@@ -191,22 +191,41 @@ class Mqtt():
         # Set up TLS support
         if self.tls_conf:
 
+            # Creating the tls context
+            if ssl is None:
+                log.error("This platform has no SSL/TLS")
+                raise ValueError("This platform has no SSL/TLS")
+
+            if not hasattr(ssl, 'SSLContext'):
+                # Require Python version that has SSL context support in standard library
+                log.error("Python 2.7.9 and 3.2 are the minimum supported versions for TLS")
+                raise ValueError("Python 2.7.9 and 3.2 are the minimum supported versions for TLS")
+
             # Validate CA certificate path
-            if self.identity.root_ca_cert:
-                if not(os.path.exists(self.identity.root_ca_cert)):
-                    log.error("Error : Wrong CA certificate path.")
-                    raise ValueError("Error : Wrong CA certificate path.")
-            else:
+            if self.identity.root_ca_cert is None and not hasattr(ssl.SSLContext, 'load_default_certs'):
                 log.error("Error : CA certificate path is missing")
                 raise ValueError("Error : CA certificate path is missing")
+            else:
+                if not(os.path.exists(self.identity.root_ca_cert)):
+                    log.error("Error : Wrong CA certificate path")
+                    raise ValueError("Error : Wrong CA certificate path")
+
+            if self.tls_conf.tls_version is None:
+                tls_version = ssl.PROTOCOL_TLSv1
+                # If the python version supports it, use highest TLS version automatically
+                if hasattr(ssl, "PROTOCOL_TLS"):
+                    tls_version = ssl.PROTOCOL_TLS
+            else:
+                tls_version = getattr(ssl, self.tls_conf.tls_version)
+            context = ssl.SSLContext(tls_version)
 
             # Validate client certificate path
             if self.identity.cert_file:
                 if os.path.exists(self.identity.cert_file):
                     client_cert_available = True
                 else:
-                    log.error("Error : Wrong client certificate path.")
-                    raise ValueError("Error : Wrong client certificate path.")
+                    log.error("Error : Wrong client certificate path")
+                    raise ValueError("Error : Wrong client certificate path")
             else:
                 client_cert_available = False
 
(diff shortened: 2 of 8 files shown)


liota/lib/transports/mqtt.py
```diff
@@ -191,22 +190,45 @@ def connect_soc(self):
...
+            if self.identity.root_ca_cert is None and not hasattr(ssl.SSLContext, 'load_default_certs'):
                 log.error("Error : CA certificate path is missing")
                 raise ValueError("Error : CA certificate path is missing")
+            else:
+                if self.identity.root_ca_cert and not (os.path.exists(self.identity.root_ca_cert)):
+                    log.error("Error : Wrong CA certificate path")
+                    raise ValueError("Error : Wrong CA certificate path")
+
+            if self.tls_conf.tls_version is None:
+                tls_version = ssl.PROTOCOL_TLSv1
```
KohliDev: TLSv1_2

liota/lib/utilities/tls_conf.py
```diff
@@ -40,7 +40,7 @@ class TLSConf:
     This class encapsulates TLS config related parameters.
     """
 
-    def __init__(self, cert_required, tls_version, cipher):
+    def __init__(self, cert_required, tls_version, cipher, crl_path=None):
```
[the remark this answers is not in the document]
KohliDev: It is as per the connection based on ca_certificate, hence it is a parameter in "tlsconf" and cannot be fixed for the gateway.

liota/lib/utilities/tls_conf.py
```diff
@@ -50,4 +50,5 @@ def __init__(self, cert_required, tls_version, cipher):
         self.cert_required = cert_required
         self.tls_version = tls_version
         self.cipher = cipher
+        self.crl_path = crl_path
```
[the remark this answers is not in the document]
KohliDev: It is a parameter required to be passed from user_code, added an example to clarify the usage.

liota/lib/transports/mqtt.py
```diff
@@ -191,22 +190,45 @@ def connect_soc(self):
         # Set up TLS support
         if self.tls_conf:
 
+            if self.identity is None:
+                log.error("Identity required to be set")
```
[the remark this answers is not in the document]
KohliDev: Removed the log messages for raise statements. I was just following the format in this file which was implemented before.

liota/lib/transports/mqtt.py
```diff
@@ -191,22 +190,45 @@ def connect_soc(self):
...
+                raise ValueError("Identity required to be set")
+
+            # Creating the tls context
+            if ssl is None:
```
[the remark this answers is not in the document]
KohliDev: added ssl import failure exception.

liota/lib/transports/mqtt.py
```diff
@@ -191,22 +190,45 @@ def connect_soc(self):
...
+                log.error("This platform has no SSL/TLS")
+                raise ValueError("This platform has no SSL/TLS")
+
+            if not hasattr(ssl, 'SSLContext'):
+                # Require Python version that has SSL context support in standard library
+                log.error("Python 2.7.9 and 3.2 are the minimum supported versions for TLS")
```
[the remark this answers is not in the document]
KohliDev: Removed this check, previously I followed the implementation of ssl library in paho mqtt hence added the above statements.

liota/lib/transports/mqtt.py
```diff
@@ -191,22 +193,35 @@ def connect_soc(self):
         # Set up TLS support
         if self.tls_conf:
 
+            if self.identity is None:
+                raise ValueError("Identity required to be set")
+
+            # Creating the tls context
+            if ssl is None:
```
gkmccready: Why fail here rather than at 38?

liota/lib/transports/mqtt.py
```diff
@@ -215,51 +230,79 @@ def connect_soc(self):
...
+            # Setting the tls context
+            self._paho_client.tls_set_context(context)
+
+            if getattr(ssl, self.tls_conf.cert_required) != ssl.CERT_NONE:
+                # Default to secure, sets context.check_hostname attribute
+                # if available
+                self._paho_client.tls_insecure_set(False)
+            else:
+                # But with ssl.CERT_NONE, we can not check_hostname
+                self._paho_client.tls_insecure_set(True)
```
gkmccready: How does CERT_NONE get set?

liota/lib/utilities/tls_conf.py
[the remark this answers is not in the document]
gkmccready: Generally you'll want a single directory on a host (er, gateway) that contains all of the CRLs that you need to check against.  You won't have a different directory of CRL(s) that you want to check for different connections; if a certificate is revoked, it's revoked for all connections trying to use it.

packages/sampleProp.conf
```diff
@@ -33,6 +33,7 @@ broker_username = "Username"
 broker_password = "Password"
 cert_required = "CERT_REQUIRED"
 tls_version = "PROTOCOL_TLSv1_2"
+crl_path = "/etc/liota/mqtt/sha2-server-g5.crl.pem"
```
gkmccready: If we're going to do a single CRL file in pem format we're going to need to document how to maintain it; there's a good chance users will need to combine CRLs from multiple CAs. /etc is often read-only; we'd be better served using /usr/lib/... or wherever as our example since this the CRL is something that will need to be updated at runtime.

liota/lib/transports/mqtt.py
```diff
@@ -337,18 +380,16 @@ def disconnect(self):
             ten_ms_count += 1
             time.sleep(0.01)
```
gkmccready: What's this weird delay loop about? Please add a comment.

liota/lib/transports/mqtt.py
KohliDev: I followed the implementation in Paho MQTT client as this piece of code (creating context) is similar to Paho MQTT internal code:
https://github.com/eclipse/paho.mqtt.python/blob/master/src/paho/mqtt/client.py#L26

I believe the reason is to complete the installation of the library even if "ssl" is missing and then raise valueError if in case the user configures tls/ssl.

liota/lib/transports/mqtt.py
KohliDev: It is passed from the user code by setting a parameter in sampleProp.conf

https://github.com/vmware/liota/blob/master/packages/sampleProp.conf#L34

liota/lib/utilities/tls_conf.py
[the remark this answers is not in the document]
KohliDev: Using the same Liotad user can publish data to AWS and IBM Bluemix. If the user is having two different CA's one for AWS and other for IBM Bluemix, I'm confused will the CA's revocation check will be done using the same CRL combined list or different?

packages/sampleProp.conf
KohliDev: I just provided an example for the user code it is not a fixed path. If we are going to combine CRL`s from multiple path then it should be placed in /usr/lib.

liota/lib/transports/mqtt.py
KohliDev: @Venkat2811  Can you please provide comment on it?

I think it is basically to complete the disconnect.

liota/lib/utilities/tls_conf.py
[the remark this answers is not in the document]
gkmccready: Generally people will combine all of the CRLs they care about in to a single file. Most CRLs are distributed in DER format, but OpenSSL wants the CRL in PEM, so the user is already taking steps to set up the CRL properly.

packages/sampleProp.conf
gkmccready: When we provide examples/samples we should try to keep them realistic, so using /etc rather than /usr for the CRL is not a great example.

Generally a sysadmin will get all the CRLs they care about, combine them and then distribute them out to the machines that need them as a single file.

liota/lib/transports/mqtt.py
gkmccready: Ah yes. Because we're still supporting non-TLS connections. Thank you.

packages/sampleProp.conf
KohliDev: I have updated the PR now the crl_path is configured using liota.conf.

KohliDev:
Also, needed the approval from @gkmccready or @secursive.

config/liota.conf
```diff
@@ -10,6 +10,9 @@ entity_file_path = /usr/lib/liota/entity
 iotcc_path = /usr/lib/liota/iotcc.json
 iotcc_load_retry = 3
 
+[CRL_PATH]
```
KohliDev: @gkmccready I have made this change after your approval. I believe it should be correct to set the conf value "blank" as this is the preferred way in python rather than passing None which will be required to be parsed as string in the code.

https://docs.python.org/3/library/configparser.html#supported-ini-file-structure
section [No Values]

gkmccready: Are "" because they've been declared but no value assigned? versus None if they're not listed at all? I was just hoping not to test ("" || None) everywhere in the code...

KohliDev: I have tested "" works. But I don`t want user to set the value "/etc/liota/mqtt/cert.cl" later as the conf file parser will not accept quotes so I have set it as blank.

gkmccready: If it works. I'm not a Python guru so I'm not 100% sure where we should be using "" vs None. I just don't like testing against both everywhere. You lost me on the quoted filename?

PR #164, opened 2017-10-17
Remove Helix Handshake Protocol and Websocket IotCC

pmasrani:


liota/dccs/iotcc.py
```diff
@@ -150,6 +133,10 @@ def on_response(msg):
 
             return RegisteredEntity(entity_obj, self, self.reg_entity_id)
 
+    def _check_version(self, json_msg):
+	if json_msg["version"] != self._version:
```
gkmccready: Recommend making version an integer and allowing yourself to support a range. A good integer version number is YYYYMMDDxx -- new versions will always be a bigger value... you don't get major.minor.dot but it works pretty well and is a fast parse and comparison.

liota/dccs/iotcc.py
```diff
@@ -71,7 +71,6 @@ def __init__(self, con):
         thread.start()
         # Wait for Subscription to be complete and then proceed to publish message
         time.sleep(0.5)
```
gkmccready: There must be a better way to do this. Sleeping is never the right choice. Why wouldn't we just block waiting for the reply? If the server is slow what ensures 500ms is enough, anyway?

liota/dccs/iotcc.py
```diff
@@ -71,7 +71,6 @@ def __init__(self, con):
...
-        self.proto = HelixProtocol(self.comms, self.comms.identity.username, self.comms.identity.password)
```
gkmccready: If we never do this how does username & password get set and used? Oh, right. We're bundled in the MQTT session now which is authenticated and secured.

liota/dccs/iotcc.py
pmasrani: Will raise a separate bug for this.

liota/dccs/iotcc.py
pmasrani: Yes, you are right.

pmasrani:
Can you please elaborate on why it is confusing?

winniex1:
I could not see your message in github, so answer you here.
In our liota request/response messages, “type” normally is message type, e.g,
create_or_find_resource_request, etc.

Now, you have a type of “UNSUPPORTED_VERSION”. Isn’t it confusing?

Winnie

liota/dccs/iotcc.py
```diff
@@ -159,7 +146,8 @@ def on_response(msg):
             try:
                 log.debug("Received msg: {0}".format(msg))
                 json_msg = json.loads(msg)
-                log.debug("Processed msg: {0}".format(json_msg["type"]))
+                log.debug("Processing msg: {0}".format(json_msg["type"]))
+                self._check_version(json_msg)
```
KohliDev: Redundant check, you are already raising the exception during the registration call if there is version mismatch the code will never reach here. It's an extra overhead to introduce version check in the current implemented model.

PR #166, opened 2017-10-20
Unit Handling

KohliDev:
1. Only set the unit if they are passed from the user code.

2. Corrected the un-registration call, the entry should be deleted from iotcc.json file only if the resource is deleted successfully.

3. Adding a check for setting system properties for device entities.

### DIFF
--- a/liota/dccs/iotcc.py
+++ b/liota/dccs/iotcc.py
@@ -79,26 +79,6 @@ class IotControlCenter(DataCenterComponent):
         # Liota internal entity file system path special for iotcc
         self.entity_file_path = self._get_file_storage_path("entity_file_path")
         self.file_ops_lock = Lock()
-        try:
-            def on_response(msg):
-                log.debug("Received msg: {0}".format(msg))
-                json_msg = json.loads(msg)
-                self.proto.on_receive(json_msg)
-                if json_msg["type"] == "connection_response":
-                    if json_msg["body"]["result"] == "succeeded":
-                        log.info("Connection verified")
-                    else:
-                        raise Exception("Helix Protocol Version mismatch")
-                else:
-                    on_response(self.recv_msg_queue.get(True, timeout))
-                    log.debug("Processed msg: {0}".format(json_msg["type"]))
-
-            # Block on Queue for not more then 300 seconds else it will raise an exception
-            on_response(self.recv_msg_queue.get(True, timeout))
-        except Exception as error:
-            self.comms.client.disconnect()
-            log.error("HelixProtocolException: " + repr(error))
-            raise Exception("HelixProtocolException")
 
     def register(self, entity_obj):
         """ Register the objects
@@ -127,8 +107,9 @@ class IotControlCenter(DataCenterComponent):
                     else:
                         log.info("Waiting for resource creation")
                         on_response(self.recv_msg_queue.get(True, timeout))
-                except:
-                    raise Exception("Exception while registering resource")
+                except Exception as err:
+                    log.exception("Exception while registering resource")
+                    raise err
 
             if entity_obj.entity_type == "EdgeSystem":
                 entity_obj.entity_type = "HelixGateway"
@@ -178,8 +159,9 @@ class IotControlCenter(DataCenterComponent):
                 else:
                     log.info("Waiting for unregistration response")
                     on_response(self.recv_msg_queue.get(True, timeout))
-            except Exception as e:
-                    raise Exception("Exception while unregistering resource: %s" % e, sys.exc_info()[2])
+            except Exception as err:
+                log.exception("Exception while unregistering resource")
+                raise err
 
         self.comms.send(json.dumps(self._unregistration(self.next_id(), entity_obj.ref_entity)))
         on_response(self.recv_msg_queue.get(True, timeout))
--- a/liota/dccs/iotcc.py
+++ b/liota/dccs/iotcc.py
@@ -90,8 +90,8 @@ class IotControlCenter(DataCenterComponent):
                     else:
                         raise Exception("Helix Protocol Version mismatch")
                 else:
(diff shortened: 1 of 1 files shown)


liota/dccs/iotcc.py
```diff
@@ -162,21 +162,19 @@ def on_response(msg):
...
         self.comms.send(json.dumps(self._unregistration(self.next_id(), entity_obj.ref_entity)))
         on_response(self.recv_msg_queue.get(True, 20))
-        if entity_obj.ref_entity.entity_type != "HelixGateway":
-            self.store_device_info(entity_obj.reg_entity_id, entity_obj.ref_entity.name,
-                                   entity_obj.ref_entity.entity_type, None, True)
-        else:
-            self.remove_reg_entity_details(entity_obj.ref_entity.name, entity_obj.reg_entity_id)
-            self.store_device_info(entity_obj.reg_entity_id, entity_obj.ref_entity.name, None, None, True)
-
-        log.info("Unregistration of resource {0} with IoTCC complete".format(entity_obj.ref_entity.name))
```
winniex1: What's the benefit of moving 172-178 to 165? In fact, they are doing the same thing. 
L179 log.info seems useful since L164 just logged unregistration with IoTCC succeeded, while L179 represents the unregistration is fully complete.

KohliDev: It is wrongly placed as we figured out in our testing: If the registration of the resource gets failed and enter the else loop at line 165. Then also the below code gets executed and entry is deleted from the json file which shouldn't be the case. Hence, the entry should be deleted from the file only once the un-registration is successfully completed for that resource and your receive the response back.

Line number 179 gets logged even if the resource un-registration is not successful it creates confusion if the resource registration is unsuccessful.

KohliDev: I have corrected the un-registration call making it similar to registration call as of now.

liota/dccs/iotcc.py
```diff
@@ -162,21 +162,20 @@ def on_response(msg):
                 log.debug("Processed msg: {0}".format(json_msg["type"]))
                 if json_msg["type"] == "remove_resource_response" and json_msg["body"]["result"] == "succeeded":
                     log.info("Unregistration of resource {0} with IoTCC succeeded".format(entity_obj.ref_entity.name))
+                    if entity_obj.ref_entity.entity_type != "HelixGateway":
+                        self.store_device_info(entity_obj.reg_entity_id, entity_obj.ref_entity.name,
+                                               entity_obj.ref_entity.entity_type, None, True)
+                    else:
+                        self.remove_reg_entity_details(entity_obj.ref_entity.name, entity_obj.reg_entity_id)
+                        self.store_device_info(entity_obj.reg_entity_id, entity_obj.ref_entity.name, None, None, True)
```
pmasrani: Isn't line 166 and 170 same? Shouldn't we take it out of if..else?

liota/dccs/iotcc.py
```diff
@@ -162,21 +162,20 @@ def on_response(msg):
...
                 else:
-                    log.info("Unregistration of resource {0} with IoTCC failed".format(entity_obj.ref_entity.name))
+                    log.info("Waiting for unregistration response")
+                    on_response(self.recv_msg_queue.get(True, 20))
```
pmasrani: Can we change the wait time to 300 seconds everywhere? Introduce a constant and use it please. It is really difficult to track different values at different places.

liota/dccs/iotcc.py
```diff
@@ -162,21 +162,20 @@ def on_response(msg):
...
             except:
                 raise Exception("Exception while unregistering resource")
```
pmasrani: Shouldn't we raise the original exception? In this way, stacktrace and original exception message will be lost.

liota/dccs/iotcc.py
```diff
@@ -162,21 +162,20 @@ def on_response(msg):
...
 
         self.comms.send(json.dumps(self._unregistration(self.next_id(), entity_obj.ref_entity)))
         on_response(self.recv_msg_queue.get(True, 20))
```
pmasrani: Ditto

liota/dccs/iotcc.py
```diff
@@ -162,21 +162,20 @@ def on_response(msg):
                 log.debug("Processed msg: {0}".format(json_msg["type"]))
```
pmasrani: Incorrect log, message is yet to be processed.

liota/dccs/iotcc.py
KohliDev: I believe the 300 seconds was added in order to handle and tackle the situation of bulk registrations in which the IoTCC takes the significant amount of time to respond. But, that`s not the case with un-registration we are introducing wait time if in case we need to remove any stale response messages which exist in the queue.

liota/dccs/iotcc.py
KohliDev: Fixed

liota/dccs/iotcc.py
KohliDev: Fixed.

liota/dccs/iotcc.py
pmasrani: 300 seconds safeguard is to ensure the client waits for the server to respond back for 5 minutes. If the server is heavily loaded, it will not respond back in 20 seconds.

KohliDev: I have changed it to 300 the only concern I have is waiting too long for the response might not be required. Hopefully, they are no stale response messages(which are not consumed) in the Queue from HelixAdapter after refactoring, which will otherwise add the significant amount of delay as they are removed from the Queue in this method.

pmasrani: I don't understand why consumption of a message, not intended as a response for this request should add "significant amount of delay". Especially, how does the values being 20 or 300 make a difference in consumption of these messages?

PR #167, opened 2017-10-20
Add Reconnect delay

KohliDev:
Adding reconnect random time so that every MQTT client connects randomly to the broker in case of disconnection.

### DIFF
--- a/liota-lite/requirements.txt
+++ b/liota-lite/requirements.txt
@@ -1,7 +1,6 @@
 aenum==1.4.5
 linux-metrics==0.1.4
 mock==2.0.0
-paho-mqtt==1.2
+paho-mqtt==1.3.1
 pint==0.7.2
 websocket-client==0.37.0
-
--- a/liota/lib/transports/mqtt.py
+++ b/liota/lib/transports/mqtt.py
@@ -35,6 +35,7 @@ import os
 import ssl
 import sys
 import time
+import random
 
 import paho.mqtt.client as paho
 
@@ -78,6 +79,12 @@ class Mqtt():
         self._connect_result_code = sys.maxsize
         self._disconnect_result_code = rc
         log.info("Disconnected with result code : {0} : {1} ".format(str(rc), paho.connack_string(rc)))
+        if self._disconnect_result_code == 0:
+            log.info("Clean disconnection")
+        else:
+            wait_time = round(random.uniform(1.0, 10.0), 1)
+            log.info("Unexpected disconnection! Reconnecting in {0} seconds".format(str(wait_time)))
+            time.sleep(wait_time)
 
     def on_message(self, client, userdata, msg):
         """
@@ -271,6 +278,7 @@ class Mqtt():
         self._paho_client.connect(host=self.url, port=self.port, keepalive=self.keep_alive)
 
         # Start network loop to handle auto-reconnect
+        self._paho_client.reconnect_delay_set(random.randint(1, 15), random.randint(150, 250))
         self._paho_client.loop_start()
         ten_ms_count = 0
         while (ten_ms_count != self._conn_disconn_timeout * 100) and (self._connect_result_code == sys.maxsize):
@@ -308,7 +316,7 @@ class Mqtt():
             log.info("Publishing Message ID : {0} with result code : {1} ".format(mess_info.mid, mess_info.rc))
             log.debug("Published Topic:{0}, Payload:{1}, QoS:{2}".format(topic, message, qos))
         except Exception:
-            log.exception("MQTT Publish exception traceback..")
+            raise Exception("MQTT Publish exception traceback..")
 
     def subscribe(self, topic, qos, callback):
         """
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,6 @@ CoAPthon==4.0.2
 aenum==1.4.5
 linux-metrics==0.1.4
 mock==2.0.0
-paho-mqtt==1.2
+paho-mqtt==1.3.1
 pint==0.7.2
 websocket-client==0.37.0
-


liota/lib/transports/mqtt.py
```diff
@@ -78,6 +79,12 @@ def on_disconnect(self, client, userdata, rc):
         self._connect_result_code = sys.maxsize
         self._disconnect_result_code = rc
         log.info("Disconnected with result code : {0} : {1} ".format(str(rc), paho.connack_string(rc)))
+        if self._disconnect_result_code == 0:
+            log.info("Clean disconnection")
+        else:
+            wait_time = round(random.uniform(1.0, 10.0), 1)
```
gkmccready: Please use random.randint(1, 10)  as you do elsewhere, no need for floating point... and please add a comment explaining why we delay and how the upper bound of 10 seconds was chosen.

liota/lib/transports/mqtt.py
```diff
@@ -271,6 +278,7 @@ def connect_soc(self):
         self._paho_client.connect(host=self.url, port=self.port, keepalive=self.keep_alive)
 
         # Start network loop to handle auto-reconnect
+        self._paho_client.reconnect_delay_set(random.randint(1, 15), random.randint(150, 250))
```
gkmccready: This is too random for us to understand the behaviour; isn't paho going to slowly move between min and max, anyway, so we should just say 1, 250? Otherwise we need a good comment explaining why we're making min & max random values.

liota/lib/transports/mqtt.py
KohliDev: I added the delay so that every mqtt client have the different random delay(different gateway) before retrying to connect, thus reducing the load on the broker. I used float trying to make it more random. Will add the comment.

liota/lib/transports/mqtt.py
KohliDev: When connection is lost, wait initially min_delay seconds and double this time every attempt. The wait is capped at max_delay as per the documentation in Paho MQTT Library.

I have added random delay so that after the max_delay limit is reached every client retries at the different max_delay time.

liota/lib/transports/mqtt.py
```diff
@@ -78,6 +79,12 @@ def on_disconnect(self, client, userdata, rc):
         self._connect_result_code = sys.maxsize
         self._disconnect_result_code = rc
         log.info("Disconnected with result code : {0} : {1} ".format(str(rc), paho.connack_string(rc)))
+        if self._disconnect_result_code == 0:
+            log.info("Clean disconnection")
+        else:
+            wait_time = round(random.uniform(1.0, 10.0), 1)
+            log.info("Unexpected disconnection! Reconnecting in {0} seconds".format(str(wait_time)))
```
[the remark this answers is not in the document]
KohliDev: The RC code is included above which will inform the user about the reason for disconnection as per MQTT. I`ll refactor the log statement.

liota/lib/transports/mqtt.py
gkmccready: Being a float and rounding to a whole number doesn't make this any more random.

liota/lib/transports/mqtt.py
gkmccready: You don't think that the various starting times of trying to make the failed connection by each individual client will be random enough? They're not all operating in-sync...

KohliDev: As per QE testing, they figured out that the MQTT broker rejects client connection as all of them try to connect together (due to a large number of DB calls for ACL authorization, it wasn`t occurring before ACL was implemented). This issue should be resolved at database front but I`m trying to add some randomness for every client as they try to re-connect.

liota/lib/transports/mqtt.py
KohliDev: I wasn't rounding to complete whole number it still had once decimal place but agreed not added much random effect.

liota/lib/transports/mqtt.py
gkmccready: Again, don't you think there will be enough randomness across clients from when they first try to connect (and fail), and how quickly every subsequent attempt occurs? (I think there is enough randomness there, especially as you back off to almost 5 minutes between retries!) This sounds like you're fixing the QE environment, and not the customer environment. I bet this is a scale test where everything starts very quickly and closely together? I agree we need to support that, but does that apply to a customer environment and should they be penalized with these long back offs?

liota/lib/transports/mqtt.py
gkmccready: Python got me. :-) But I still disagree with using floats when they are not necessary.

liota-lite/requirements.txt
```diff
@@ -1,7 +1,5 @@
 aenum==1.4.5
 linux-metrics==0.1.4
-mock==2.0.0
-paho-mqtt==1.2
+paho-mqtt==1.3.1
```
pmasrani: I think there is something really wrong with the way liota-lite is designed. There should be a way to have a common file structure to describe the requirements and setup.py, with a flag being passed. This is a developer's nightmare to keep files in sync for liota and liota-lite.

liota/lib/transports/mqtt.py
```diff
@@ -78,6 +79,14 @@ def on_disconnect(self, client, userdata, rc):
...
         self._disconnect_result_code = rc
         log.info("Disconnected with result code : {0} : {1} ".format(str(rc), paho.connack_string(rc)))
+        if self._disconnect_result_code == 0:
+            log.info("Clean disconnection")
+        else:
+            # Added a random sleep so that every client tries out reconnection to broker at different
+            # time to reduce load
+            wait_time = random.randint(1, 5)
+            log.info("Unexpected disconnection! Reconnecting in {0:d} seconds".format(wait_time))
+            time.sleep(wait_time)
```
pmasrani: As per comments of this function, it can be called by a client disconnection as well. Why should we wait here if it is a client initiated disconnection?

liota/lib/transports/mqtt.py
```diff
@@ -78,6 +79,14 @@ def on_disconnect(self, client, userdata, rc):
         self._connect_result_code = sys.maxsize
         self._disconnect_result_code = rc
         log.info("Disconnected with result code : {0} : {1} ".format(str(rc), paho.connack_string(rc)))
+        if self._disconnect_result_code == 0:
+            log.info("Clean disconnection")
+        else:
+            # Added a random sleep so that every client tries out reconnection to broker at different
+            # time to reduce load
+            wait_time = random.randint(1, 5)
```
pmasrani: Why do we have two random functions, one with a wait_time here and another at line 283? Can't we be random enough just with one single function? Or at least have them in the same place?

liota/lib/transports/mqtt.py
```diff
@@ -271,6 +280,7 @@ def connect_soc(self):
         self._paho_client.connect(host=self.url, port=self.port, keepalive=self.keep_alive)
 
         # Start network loop to handle auto-reconnect
+        self._paho_client.reconnect_delay_set(random.randint(1, 15), random.randint(150, 250))
```
pmasrani: Why these numbers where chosen? How many clients did we try out with? Are we sure this will succeed for thousands of client?
We should make these parameters user configurable so that QE can test and set it to correct values, instead of some randomly chosen numbers.

liota-lite/requirements.txt
pmasrani: You can raise a separate bug to fix this later, as it is not a concern related to this CL.

KohliDev: I don't believe this is required to be fixed as this approach is given to us by Open Source team.

liota/lib/transports/mqtt.py
KohliDev: This case is handled in the "if" loop above.

liota/lib/transports/mqtt.py
KohliDev: I agree that's the reason I have asked for completing testing report from QE pointing out if this is the main cause. I'll remove the reconnect_delay_set temporarily and just have a random sleep if the disconnection happens unexpectedly.

KohliDev:
We are putting this PR on hold as we need concrete reports and exact reason if the MQTT Broker is rejecting connection due to various clients trying at the same time as liota mqtt client has enough randomness before retrying the connection.

Venkat2811:
Please find Randomized exponential back implementation here: https://github.com/diwanc/liota/commit/a7f229ca88cb3dd75a6987798f0d56897d4d9f8c

PR #168, opened 2017-10-20
Package Manager and Device Discovery improvement

winniex1:
1) capture PackageClass instance creation exception to avoid blocking
    package manager thread continous work
    2) display package operation command execution result into stdout,
    e.g., (load, reload, update, unload, delete etc.)
    3) add support for querying a package status (loaded or not) by using
    command of "stat pkg package_name"
    4) correct package manager and device discovery log message format


liota/core/package_manager.py
```diff
@@ -426,9 +448,15 @@ def run(self):
                         continue
                     file_name = msg[1]
                     if command == "unload":
-                        self._package_unload(file_name)
+                        if (self._package_unload(file_name) == True):
+                            print "succeeded to unload:", file_name
+                        else:
+                            print "failed to unload:", file_name
                     elif command == "delete":
```
pmasrani: Does this delete the package?

liota/core/package_manager.py
```diff
@@ -252,24 +252,22 @@ def __init__(self, name=None):
...
     def _cmd_handler_list(self, parameter):
         if parameter == "packages" or parameter == "pkg":
-            log.warning("List of packages - \n\t%s"
-                        % "\n\t".join(sorted(
-                            self._packages_loaded.keys()
-                        ))
-                        )
+            loaded_list = sorted(self._packages_loaded.keys())
+            log.warning("List of packages - \t%s" % "\t".join(loaded_list))
+            print "loaded packages:", ",".join(loaded_list)
```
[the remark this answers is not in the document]
winniex1: I have discussed with Arun in email threads. Our package manager uses a loop thread to keep reading commands from command message queue, and could not return. Therefore, I could not pass a command execution status code back. Arun agrees that to get status from stdout, and in fact, he just wanted to know the execution status.

liota/core/package_manager.py
```diff
@@ -289,7 +287,19 @@ def _cmd_handler_list(self, parameter):
     # This method is used to handle statistical commands
 
     def _cmd_handler_stat(self, parameter):
-        if parameter == "metrics" or parameter == "met":
+        if parameter[0] == "package" or parameter[0] == "pkg":
+            if (len(parameter) != 2):
+                log.warning("package name is not specified")
```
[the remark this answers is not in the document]
winniex1: In our design, we only accept 1 package status query, not a list of package. Therefore, only the first parameter will be processed, i.e., in "stat pkg package_name1 package_name2", only package_name1 will be processed.
ok, I can log this out to remind user only 1 package will be processed.

liota/core/package_manager.py
```diff
@@ -289,7 +287,19 @@ def _cmd_handler_list(self, parameter):
     # This method is used to handle statistical commands
 
     def _cmd_handler_stat(self, parameter):
-        if parameter == "metrics" or parameter == "met":
+        if parameter[0] == "package" or parameter[0] == "pkg":
```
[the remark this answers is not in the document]
winniex1: before call _cmd_handler_stat(), we have checked and guaranteed there will be at least 1 parameter, which can be "met, pkg, etc..". So, we can always use parameter[0] rather than parameter to get "pkg". After this, I have sanity check to guarantee that after "pkg", we have more package_names as parameters to process.

liota/core/package_manager.py
winniex1: Yes, the "delete" cmd will delete the package in file system.

liota/core/package_manager.py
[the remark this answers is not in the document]
pmasrani: I meant, shouldn't the name of the variable be "parameters" instead of parameter? It seems incorrect to call it parameter, since it is multiple.

liota/core/package_manager.py
```diff
@@ -694,8 +722,12 @@ def _package_load(self, file_name, checksum=None, autoload_flag=False,
 
         klass = getattr(module_loaded, "PackageClass")
         package_record = PackageRecord(file_name)
-        if not package_record.set_instance(klass()):
-            log.error("Unexpected failure initializing package")
+        try:  # Create instance
```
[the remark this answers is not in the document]
winniex1: This is a fix for "liotad becomes unresponsive (DoS) when an erroneous package is load".
https://jira-euc.eng.vmware.com/jira/browse/IOT-2111

Inside the erroneous package, there is only run(), while no clean_up(). Then, the PackageClass instance creation will fail since clean_up() only has abstract class. An system exception will be raised when we try to run .set_instance(), so without "try..catch", nobody deals with the .set_instance() exception which will cause no further command could be input and processed.  log.error("Unexpected failure initializing package") cannot take effect.

In a word, we have to explicitly use "try..except" for .set_instance().

liota/core/package_manager.py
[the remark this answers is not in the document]
winniex1: in fact, so far, only "stat pkg" has further parameter, "stat met/th/col" will have no further parameter. But because they will share the same _cmd_handler_stat(), I have to use parameter[0] to process for all of them.
Yes, i agree even in above case, parameters will be more suitable as parameter. Will change it.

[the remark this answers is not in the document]
winniex1: done

liota/core/package_manager.py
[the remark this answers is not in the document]
winniex1: done

liota/core/package_manager.py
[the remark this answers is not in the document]
pmasrani: What about the non-pulse-agent users? I don't think it is a good idea to send output to stdout, in a non-blocking format. It will just pop out on the shell, when they least expect.

Ideally, there should be a two-way channel, not a unidirectional pipe between liotapkg and liotad.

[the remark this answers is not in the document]
pmasrani: Also, if package_manager is printing these values to stdout, how will these values reach the Pulse Agent? Pulse Agent might be expecting it at the time of calling liotapkg.sh, correct?

liota/core/package_manager.py
```diff
@@ -288,8 +288,20 @@ def _cmd_handler_list(self, parameter):
     #-----------------------------------------------------------------------
     # This method is used to handle statistical commands
 
-    def _cmd_handler_stat(self, parameter):
-        if parameter == "metrics" or parameter == "met":
+    def _cmd_handler_stat(self, parameters):
+        if parameters[0] == "package" or parameters[0] == "pkg":
```
pmasrani: If we are just logging it, is the command useful? Anyway we are getting this information from the logs by list pkg. I think the  purpose was to get the output to Pulse Agent, which is not met.

liota/core/package_manager.py
```diff
@@ -439,10 +451,10 @@ def run(self):
                     self._cmd_handler_list(msg[1])
             elif command == "stat":
                 with package_lock:
-                    if len(msg) != 2:
+                    if len(msg) < 2:
```
pmasrani: What about > 2?

packages/README.md
```diff
@@ -0,0 +1,107 @@
+#Liota Package Manager
```
pmasrani: There is already a Readme under liotad folder. https://github.com/vmware/liota/blob/fe00dc066feeb3eb9508bb02780aad5b4e0a8ad8/packages/liotad/README.md

If you want to move it here, we better delete the old one. It will be difficult to manage consistencies of multiple Readme files.

packages/liotad/liotapkg.sh
```diff
@@ -65,5 +65,4 @@ then
 fi
 
 # Echo to named pipe
-echo "Pipe file: $package_messenger_pipe" >&2
```
pmasrani: If we are removing this, we should have some response for the user. Something like "Command queued for execution"

liota/core/package_manager.py
winniex1: This command can let u directly query whether that pkg is loaded or not. It's more straightforward.

packages/README.md
winniex1: done

packages/liotad/liotapkg.sh
winniex1: I am ok with either way. In fact, have you ever seen you linux cmd shows "have put cmd xxx into named pipe"? It seems you are already familiar with current style, so I put it back.

liota/core/package_manager.py
winniex1: It was covered inside _cmd_handler_stat()

packages/liotad/liotapkg.sh
pmasrani: In fact, i have never seen a linux command which works on a pipe and doesn't give back the output! :)

Until we start returning the output, don't you think it is best to notify the user that the command is queued?

liota/core/package_manager.py
pmasrani: Ok, considering that once we start returning the output to the user who runs liotapkg.sh some time in future, this command will be helpful.

packages/liotad/liotapkg.sh
winniex1: You can check package manager thread directly to see why the current design doesn't give back the output. Anyway, I have put the echo message back.

liota/core/package_manager.py
```diff
@@ -288,8 +288,23 @@ def _cmd_handler_list(self, parameter):
...
+                log.warning("Only 1 package name will be taken and processed")
+            query_pkg = parameters[1]
+            if query_pkg in self._packages_loaded.keys():
+                log.info("packages {0} is loaded".format(query_pkg))
+            else:
+                log.info("packages {0} is not loaded".format(query_pkg))
+            return
+        if len(parameters) != 1:
+            log.warning("Invalid format of stat command: %s" % parameters[0])
+            continue
```
pmasrani: If it is an invalid format, why do we continue? Can't we just return?

packages/liotad/liotapkg.sh
pmasrani: The concern was never my familiarity with the echo or package manager design. It was all to give back a more useful message to the user, which will make sense to him.

liota/core/package_manager.py
winniex1: returned.

PR #169, opened 2017-10-20
Re-subscribe mechanism

KohliDev:
1. MQTT Topic re-subscribe mechanism on reconnect

2. Refactoring MQTT, MQTT_DCC_COMMS

3. Unsubscribe implementation in MQTT

4. Updated comments and README

### DIFF
--- a/liota/dcc_comms/mqtt_dcc_comms.py
+++ b/liota/dcc_comms/mqtt_dcc_comms.py
@@ -74,16 +74,17 @@ class MqttDccComms(DCCComms):
 
         self.client_id = client_id
 
+        if self.client_id is None or self.client_id == "":
+            #  local_uuid generated will be the client ID
+            self.client_id = systemUUID().get_uuid(edge_system_name)
+            log.info("generated local uuid will be the client ID")
+        else:
+            log.info("Client ID is provided by user")
+
         if mqtt_msg_attr is None:
             #  pub-topic and sub-topic will be auto-generated
             log.info("pub-topic and sub-topic is auto-generated")
             self.msg_attr = MqttMessagingAttributes(edge_system_name)
-            if self.client_id is None or self.client_id == "":
-                #  local_uuid generated will be the client ID
-                self.client_id = systemUUID().get_uuid(edge_system_name)
-                log.info("generated local uuid will be the client ID")
-            else:
-                log.info("Client ID is provided by user")
         elif isinstance(mqtt_msg_attr, MqttMessagingAttributes):
             log.info("User configured pub-topic and sub-topic")
             self.msg_attr = mqtt_msg_attr
--- a/liota/dcc_comms/mqtt_dcc_comms.py
+++ b/liota/dcc_comms/mqtt_dcc_comms.py
@@ -46,7 +46,7 @@ class MqttDccComms(DCCComms):
     """
 
     def __init__(self, edge_system_name, url, port, identity=None, tls_conf=None, qos_details=None,
-                 client_id="", clean_session=False, protocol="MQTTv311", transport="tcp", keep_alive=60,
+                 client_id=None, clean_session=False, protocol="MQTTv311", transport="tcp", keep_alive=60,
                  mqtt_msg_attr=None, enable_authentication=False, conn_disconn_timeout=10):
 
         """
@@ -74,19 +74,19 @@ class MqttDccComms(DCCComms):
 
         self.client_id = client_id
 
-        if self.client_id is None or self.client_id == "":
+        if self.client_id is None:
             #  local_uuid generated will be the client ID
             self.client_id = systemUUID().get_uuid(edge_system_name)
-            log.info("generated local uuid will be the client ID")
+            log.debug("Auto-Generated local uuid will be the client ID")
         else:
-            log.info("Client ID is provided by user")
+            log.debug("Client ID is provided by user")
 
         if mqtt_msg_attr is None:
             #  pub-topic and sub-topic will be auto-generated
-            log.info("pub-topic and sub-topic is auto-generated")
+            log.debug("Pub-topic and Sub-topic are auto-generated")
             self.msg_attr = MqttMessagingAttributes(edge_system_name)
         elif isinstance(mqtt_msg_attr, MqttMessagingAttributes):
-            log.info("User configured pub-topic and sub-topic")
+            log.debug("User configured pub-topic and sub-topic")
             self.msg_attr = mqtt_msg_attr
         else:
             log.error("mqtt_mess_attr should either be None or of type MqttMessagingAttributes")
--- a/liota/dcc_comms/mqtt_dcc_comms.py
+++ b/liota/dcc_comms/mqtt_dcc_comms.py
@@ -77,9 +77,9 @@ class MqttDccComms(DCCComms):
         if self.client_id is None:
             #  local_uuid generated will be the client ID
             self.client_id = systemUUID().get_uuid(edge_system_name)
-            log.debug("Auto-Generated local uuid will be the client ID")
+            log.debug("Auto-Generated local uuid will be the client ID {0}".format(self.client_id))
         else:
-            log.debug("Client ID is provided by user")
+            log.debug("Client ID is provided by user {0}".format(self.client_id))
 
         if mqtt_msg_attr is None:
             #  pub-topic and sub-topic will be auto-generated
--- a/liota/lib/transports/mqtt.py
+++ b/liota/lib/transports/mqtt.py
@@ -357,8 +357,7 @@ class Mqtt():
         :return:
         """
         try:
-            if topic not in self.sub_dict:
-                self.sub_dict.setdefault(topic, [qos, callback])
+            self.sub_dict.setdefault(topic, [qos, callback])
             subscribe_response = self._paho_client.subscribe(topic, qos)
             self._paho_client.message_callback_add(topic, callback)
             log.info("Topic subscribed with information: " + str(subscribe_response))
@@ -373,9 +372,9 @@ class Mqtt():
         :return:
         """
         try:
-            if topic in self.sub_dict:
-                self.sub_dict.pop(topic)
+            self.sub_dict.pop(topic, None)
             unsubscribe_response = self._paho_client.unsubscribe(topic)
+            self._paho_client.message_callback_remove(topic)
             log.info("Topic unsubscribed with information: " + str(unsubscribe_response))
         except Exception:
             log.exception("MQTT unsubscribe exception traceback..")
--- a/liota/lib/transports/mqtt.py
+++ b/liota/lib/transports/mqtt.py
@@ -51,20 +51,6 @@ class Mqtt():
     MQTT Transport implementation for LIOTA. It internally uses Python Paho library.
     """
 
-    def on_connect(self, client, userdata, flags, rc):
-        """
-        Invoked on successful connection to a broker after connection request
-
-        :param client: The client instance for this callback
-        :param userdata: The private user data as set in Client() or userdata_set()
-        :param flags: Response flags sent by the broker
-        :param rc: The connection result
-        :return:
-        """
-        self._connect_result_code = rc
-        self._disconnect_result_code = sys.maxsize
-        log.info("Connected with result code : {0} : {1} ".format(str(rc), paho.connack_string(rc)))
-
     def on_disconnect(self, client, userdata, rc):
         """
         Invoked when disconnected from broker.  Two scenarios are possible:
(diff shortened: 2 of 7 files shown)


KohliDev:
Adding @Venkat2811 in order to review the refactored code of MQTT.

liota/dcc_comms/mqtt_dcc_comms.py
```diff
@@ -74,16 +74,17 @@ def __init__(self, edge_system_name, url, port, identity=None, tls_conf=None, qo
 
         self.client_id = client_id
 
+        if self.client_id is None or self.client_id == "":
```
gkmccready: We should get consistent between None and ""; None makes sense for it not being set, and "" should be an invalid value.

liota/dcc_comms/mqtt_dcc_comms.py
```diff
@@ -74,16 +74,17 @@ def __init__(self, edge_system_name, url, port, identity=None, tls_conf=None, qo
...
+            #  local_uuid generated will be the client ID
+            self.client_id = systemUUID().get_uuid(edge_system_name)
+            log.info("generated local uuid will be the client ID")
+        else:
+            log.info("Client ID is provided by user")
```
gkmccready: Are these "info" or "debug" level? Especially if we're not recording what the clientID provided was?

liota/lib/transports/mqtt.py
```diff
@@ -165,24 +151,41 @@ def __init__(self, url, port, identity=None, tls_conf=None, qos_details=None, cl
         if self.clean_session:
             # If user passes client_id, it'll be used.  Otherwise, it is left to the underlying paho
             # to generate random client_id
-            self._paho_client = paho.Client(self.client_id, self.clean_session, self.userdata,
-                                            protocol=getattr(paho, self.protocol), transport=self.transport)
             log.info("clean_session is set to True")
```
gkmccready: info or debug level logging? why do we only log about clean_session?

liota/lib/transports/mqtt.py
```diff
@@ -165,24 +151,41 @@ def __init__(self, url, port, identity=None, tls_conf=None, qos_details=None, cl
...
+        :param client: The client instance for this callback
+        :param userdata: The private user data as set in Client() or userdata_set()
+        :param flags: Response flags sent by the broker
+        :param rc: The connection result
+        :return:
+        """
+        self._connect_result_code = rc
+        self._disconnect_result_code = sys.maxsize
+        log.info("Connected with result code : {0} : {1} ".format(str(rc), paho.connack_string(rc)))
+        if self.sub_list:
```
gkmccready: This if is unnecessary. If sub_list is empty the for loop just ends.

liota/lib/transports/mqtt.py
```diff
@@ -165,24 +151,41 @@ def __init__(self, url, port, identity=None, tls_conf=None, qos_details=None, cl
...
+            for sub_values in self.sub_list:
```
gkmccready: Please change this to:

for [topic, qos, callback] in self.sub_list:

And the appropriate changes in the following lines instead of using sub_values[x].

liota/lib/transports/mqtt.py
```diff
@@ -320,12 +323,27 @@ def subscribe(self, topic, qos, callback):
         :return:
         """
         try:
+            if [topic,qos,callback] not in self.sub_list:
```
gkmccready: In what situations would we try to add an additional subscription for the same thing? And what happens if we subscribe to the same topic with a different qos or callback? Given unsubscribe only cares about the topic, I'm assuming whatever we subscribe with should only care about the topic, too? Or otherwise overwrite the old qos and callback?

packages/user_packages/iotcc_mqtt.py
```diff
@@ -89,6 +89,7 @@ def run(self, registry):
             print "EdgeSystem registration to IOTCC failed"
 
     def clean_up(self):
-        # Unregister the edge system
+        # Unregister the edge system on package unload
+        # Kindly include the edge system un-register call for the pulse usecase
```
gkmccready: This doesn't seem specific to Pulse?

liota/lib/transports/mqtt.py
```diff
@@ -320,12 +323,27 @@ def subscribe(self, topic, qos, callback):
...
+                self.sub_list.append([topic,qos,callback])
             subscribe_response = self._paho_client.subscribe(topic, qos)
             self._paho_client.message_callback_add(topic, callback)
             log.info("Topic subscribed with information: " + str(subscribe_response))
         except Exception:
             log.exception("MQTT subscribe exception traceback..")
 
+    def unsubscribe(self, topic):
```
gkmccready: Does anything actually call this method?

liota/dcc_comms/mqtt_dcc_comms.py
KohliDev: Fixed it. There was confusion previously if the unique client_id will be generated if value is set to "None", I verified it works.

liota/dcc_comms/mqtt_dcc_comms.py
KohliDev: Changed them to Debug log as it makes more sense to be in Debug mode.

liota/lib/transports/mqtt.py
KohliDev: Removed the logging about clean_session as it is not required. And provided a comment in README.

liota/lib/transports/mqtt.py
KohliDev: Removed it.

liota/lib/transports/mqtt.py
KohliDev: Done

liota/lib/transports/mqtt.py
KohliDev: This is an implementation which I have done in MQTT transports library which was missing previously in Liota. This method can be used by some other DCC`s in order to unsubscribe.

packages/user_packages/iotcc_mqtt.py
KohliDev: Updated the comment.

liota/lib/transports/mqtt.py
KohliDev: I did some testing and figured out that the Paho Python mqtt client subscribes to only one topic and basically it should be overriding the QoS and callback if the new subscription arrives for the same topic as I didn't receive duplicate messages. This check I have included keeping that in mind as some desktop mqtt client allow to subscribe to multiple topics and the messages get duplicated.

gkmccready: Hm. But then sub_list is not an accurate representation of what is subscribed, right?

Subscribe(foo, qos1, myCallback)
Subscribe(foo, qos2, myOtherCallback)
Unsubscribe(foo)

What's in the list and what is the client subscribed to now?

KohliDev: I have already opened an issue at Paho Python MQTT Github project to understand the behavior correctly:

https://github.com/eclipse/paho.mqtt.python/issues/241

As per my testing as library subscribes to only one topic with QoS & Callback, I plan to move from sub_list to python dictionary where the key will be the topic & values will be [Qos, Callback] and if the topic is unsubscribed it will be removed from the map. I`ll wait for a day for the final confirmation from the paho developers on the library behavior before making this change.

pmasrani: In message_callback_add(), Paho MQTT client is adding the new callback and doesn't keep a list of all callbacks registered till now. (https://github.com/eclipse/paho.mqtt.python/blob/master/src/paho/mqtt/client.py#L1735)

I think it is better to remove the check.

liota/lib/transports/mqtt.py
```diff
@@ -320,12 +314,27 @@ def subscribe(self, topic, qos, callback):
...
+        """
+        Unsubscribes to a topic
+
+        :param topic: Unsubscribe topic
+        :return:
+        """
+        try:
+            unsubscribe_response = self._paho_client.unsubscribe(topic)
```
pmasrani: I thought the whole purpose of implementing unsubscribe() was to remove it from sub_list?

liota/lib/transports/mqtt.py
KohliDev: Implemented a python dictionary rather than list it adds the topic,QoS & Callback on_subscribe() method and removes them on_unsubscribe()

liota/lib/transports/mqtt.py
```diff
@@ -320,12 +314,29 @@ def subscribe(self, topic, qos, callback):
         :return:
         """
         try:
+            if topic not in self.sub_dict:
```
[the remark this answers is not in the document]
KohliDev: The if loop should go away, but the map should be updated if new QoS and callback is added for the same topic.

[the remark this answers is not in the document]
pmasrani: Yes, of course! :)
when i see, "remove this check", i obviously mean only to remove the if statement, not the statements inside it.

BTW, if is not a loop, it is a statement.

liota/dcc_comms/mqtt_dcc_comms.py
```diff
@@ -74,18 +74,19 @@ def __init__(self, edge_system_name, url, port, identity=None, tls_conf=None, qo
 
         self.client_id = client_id
 
+        if self.client_id is None:
+            #  local_uuid generated will be the client ID
+            self.client_id = systemUUID().get_uuid(edge_system_name)
+            log.debug("Auto-Generated local uuid will be the client ID")
```
gkmccready: Line 80 & 83: Should we also log the clientID if we're going to log at all?

liota/dcc_comms/mqtt_dcc_comms.py
```diff
@@ -74,18 +74,19 @@ def __init__(self, edge_system_name, url, port, identity=None, tls_conf=None, qo
...
+        else:
+            log.debug("Client ID is provided by user")
+
         if mqtt_msg_attr is None:
             #  pub-topic and sub-topic will be auto-generated
-            log.info("pub-topic and sub-topic is auto-generated")
+            log.debug("Pub-topic and Sub-topic are auto-generated")
```
gkmccready: s/are/will be/ ? (We could/should log them?)

liota/dcc_comms/mqtt_dcc_comms.py
```diff
@@ -74,18 +74,19 @@ def __init__(self, edge_system_name, url, port, identity=None, tls_conf=None, qo
...
             self.msg_attr = MqttMessagingAttributes(edge_system_name)
-            if self.client_id is None or self.client_id == "":
-                #  local_uuid generated will be the client ID
-                self.client_id = systemUUID().get_uuid(edge_system_name)
-                log.info("generated local uuid will be the client ID")
-            else:
-                log.info("Client ID is provided by user")
         elif isinstance(mqtt_msg_attr, MqttMessagingAttributes):
-            log.info("User configured pub-topic and sub-topic")
+            log.debug("User configured pub-topic and sub-topic")
```
gkmccready: Should we log the topics? I think so if we're going to log this at all.

liota/lib/transports/mqtt.py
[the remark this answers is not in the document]
KohliDev: Thanks for correcting it and the explanation.

liota/lib/transports/mqtt.py
```diff
@@ -320,12 +314,29 @@ def subscribe(self, topic, qos, callback):
...
+        """
+        Unsubscribes to a topic
+
+        :param topic: Unsubscribe topic
+        :return:
+        """
+        try:
+            if topic in self.sub_dict:
+                self.sub_dict.pop(topic)
+            unsubscribe_response = self._paho_client.unsubscribe(topic)
```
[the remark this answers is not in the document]
KohliDev: Fixed

liota/lib/transports/mqtt.py
KohliDev: It has already been implemented post Glen`s comment.

liota/dcc_comms/mqtt_dcc_comms.py
KohliDev: Fixed

liota/dcc_comms/mqtt_dcc_comms.py
KohliDev: It is logged in MqttMessagingAttributes

liota/dcc_comms/mqtt_dcc_comms.py
KohliDev: It is logged in MqttMessagingAttributes

PR #170, opened 2017-10-20
Handling Exception

KohliDev:
1. Raise exception from transports mqtt layer.

2. Improve & Correct metric_handler and registered_metric to handle exceptions properly.

### DIFF
--- a/liota/entities/metrics/registered_metric.py
+++ b/liota/entities/metrics/registered_metric.py
@@ -121,7 +121,10 @@ class RegisteredMetric(RegisteredEntity):
         if not self.values:
             # No values measured since last report_data
             return True
-        self.ref_dcc.publish(self)
+        try:
+            self.ref_dcc.publish(self)
+        except Exception:
+            log.error("Exception while publishing message", exc_info=True)
 
     def __str__(self, *args, **kwargs):
         return str(self.ref_entity.name) + ":" + str(self._next_run_time)
--- a/liota/entities/metrics/registered_metric.py
+++ b/liota/entities/metrics/registered_metric.py
@@ -123,7 +123,7 @@ class RegisteredMetric(RegisteredEntity):
             return True
         try:
             self.ref_dcc.publish(self)
-        except PublishFailure:
+        except Exception:
             log.error("Exception while publishing message", exc_info=True)
 
     def __str__(self, *args, **kwargs):
@@ -136,5 +136,3 @@ class RegisteredMetric(RegisteredEntity):
             return -1
         return cmp(self._next_run_time, other._next_run_time)
 
-class PublishFailure(Exception):
-    pass
--- a/liota/entities/metrics/registered_metric.py
+++ b/liota/entities/metrics/registered_metric.py
@@ -135,4 +135,3 @@ class RegisteredMetric(RegisteredEntity):
         if not isinstance(other, RegisteredMetric):
             return -1
         return cmp(self._next_run_time, other._next_run_time)
-
--- a/liota/entities/metrics/registered_metric.py
+++ b/liota/entities/metrics/registered_metric.py
@@ -123,7 +123,7 @@ class RegisteredMetric(RegisteredEntity):
             return True
         try:
             self.ref_dcc.publish(self)
-        except Exception:
+        except PublishFailure:
             log.error("Exception while publishing message", exc_info=True)
 
     def __str__(self, *args, **kwargs):
@@ -135,3 +135,6 @@ class RegisteredMetric(RegisteredEntity):
         if not isinstance(other, RegisteredMetric):
             return -1
         return cmp(self._next_run_time, other._next_run_time)
+
+class PublishFailure(Exception):
+    pass
--- a/liota/lib/transports/mqtt.py
+++ b/liota/lib/transports/mqtt.py
@@ -346,7 +346,7 @@ class Mqtt():
             if mess_info.rc == 0:
                 log.debug("Published Topic:{0}, Payload:{1}, QoS:{2}".format(topic, message, qos))
             else:
-                raise Exception("MQTT Publish exception:{0} result code".format(mess_info.rc))
+                raise PublishFailure("MQTT Publish exception:{0} result code".format(mess_info.rc))
         except Exception:
             raise Exception("MQTT Publish exception")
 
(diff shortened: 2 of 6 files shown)


packages/user_packages/iotcc_mqtt.py
```diff
@@ -79,7 +80,17 @@ def run(self, registry):
             self.iotcc.set_system_properties(self.iotcc_edge_system, registry.get("system_properties"))
             # Set the properties for edge system as key:value pair, you can also set the location
             # by passing the latitude and longitude as a property in the user package
-            self.iotcc.set_properties(self.iotcc_edge_system,{"key1": "value1", "key2": "value2"})
+            # If the set_properties or register call fails due to DCC_Comms Publish exception
+            # the optional retry mechanism can be implemented in the following way
+            attempts = 0
+            while attempts < 3:
```
gkmccready: This doesn't seem optional, and isn't this going to be an infinite loop when the call succeeds since we don't break out of the while on success? If it fails 3 times the loop will break when the condition is met, though... and we don't log that failure?

packages/user_packages/iotcc_mqtt_edge_system_stats.py
```diff
@@ -65,7 +65,14 @@ def read_cpu_utilization(sample_duration_sec=1):
 
 
 def read_disk_usage_stats():
-    return round(disk_stat.disk_reads_writes(disk_name)[0], 2)
+    # If the device raises an intermittent exception during metric collection process it will be required
+    # to be handled in the user code otherwise collection process will be stopped for that metric.
+    # If the None value is returned by UDM then metric value for that particular collector instance won't be published
+    try:
+        disk_stat_value = round(disk_stat.disk_reads_writes(disk_name)[0], 2)
```
gkmccready: round() here confuses me; reads and writes can only be whole numbers?

packages/user_packages/iotcc_mqtt.py
```diff
@@ -79,7 +80,17 @@ def run(self, registry):
...
+                try:
+                    # Register edge system (gateway)
+                    self.iotcc.set_properties(self.iotcc_edge_system, {"key1": "value1", "key2": "value2"})
+                except Exception:
+                    attempts += 1
+                    # The sleep time before re-trying depends on broker restart
```
gkmccready: I don't understand this comment since the sleep time is a constant; and don't we need all of register/set_system_properties/set_properties to succeed to function properly?

packages/user_packages/README.md
```diff
@@ -16,8 +16,12 @@ System Properties which are pre-loaded in the registry during installation are r
 all the devices will internally inherit the the system properties from the registered edge system.
 The properties for the edge system can be set as 'key:value' pair, you can also set the location by passing the
 'latitude:value' and 'longitude:value' as properties in the user package.
+Optional Mechanism: If the set_properties or register call fails due to DCC_Comms publish exception
```
gkmccready: The retry mechanism does not appear optional. It always happens and is hard-coded to 3 attempts? (And a 5 second sleep between attempts?)

packages/user_packages/iotcc_mqtt_edge_system_stats.py
```diff
@@ -65,7 +65,14 @@ def read_cpu_utilization(sample_duration_sec=1):
 
 
 def read_disk_usage_stats():
-    return round(disk_stat.disk_reads_writes(disk_name)[0], 2)
+    # If the device raises an intermittent exception during metric collection process it will be required
```
pmasrani: If the device raises an intermittent exception during metric collection process, user code will be required to handled it. Otherwise, if an Exception is thrown from UDM, collection process will be stopped for that metric.
 
If UDM returns a None value, then metric value for that particular collector instance won't be published.

packages/user_packages/iotcc_mqtt_edge_system_stats.py
```diff
@@ -65,7 +65,14 @@ def read_cpu_utilization(sample_duration_sec=1):
 
 
 def read_disk_usage_stats():
```
pmasrani: Can you also add an example, where say after publishing 3-4 values, an exception is thrown, to show it's collection getting stopped?

liota/entities/metrics/registered_metric.py
```diff
@@ -121,7 +121,10 @@ def send_data(self):
         if not self.values:
             # No values measured since last report_data
             return True
-        self.ref_dcc.publish(self)
+        try:
+            self.ref_dcc.publish(self)
+        except Exception:
+            log.error("Exception while publishing message", exc_info=True)
```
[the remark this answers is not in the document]
gkmccready: I think exc_info=True makes log.error behave like log.exception

liota/lib/transports/mqtt.py
```diff
@@ -302,13 +302,12 @@ def publish(self, topic, message, qos, retain=False):
         :param retain: Message to be retained or not
         :return:
         """
-        # TODO: Retry logic to be designed
-        try:
-            mess_info = self._paho_client.publish(topic, message, qos, retain)
-            log.info("Publishing Message ID : {0} with result code : {1} ".format(mess_info.mid, mess_info.rc))
+        mess_info = self._paho_client.publish(topic, message, qos, retain)
+        log.debug("Publishing Message ID : {0} with result code : {1} ".format(mess_info.mid, mess_info.rc))
```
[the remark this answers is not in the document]
gkmccready: Echoing your comment about topic, etc, here, too. :-)

liota/entities/metrics/registered_metric.py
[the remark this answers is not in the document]
pmasrani: Didn't notice exc_info=True at all! 

I think exc_info=True should be used if we are logging at Warn/Info level. Here, we are anyway logging at Error level. log.exception() will make it easier to read.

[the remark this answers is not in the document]
KohliDev: Is there any official documentation which says that?

It does exactly the same thing and also gives the flexibility to change the logging level required in future.

packages/user_packages/iotcc_mqtt.py
KohliDev: We have just provided an example to liota developer if required he can change the wait time or even remove it. Added elaborated comments.

packages/user_packages/iotcc_mqtt_edge_system_stats.py
KohliDev: Fixed

PR #177, opened 2017-10-30
Package Manager Improvement:

winniex1:
modify liotapkg.sh to cover a case where multiple liotad processes run


winniex1:
not to block Vaibhav's work

PR #181, opened 2017-11-10
Validating names and properties against a whitelist character set supported by IoTCC

arunv10:
IoTCC expects names and properties to have only characters like - alphabets, digits, underscore, hyphen , dot , space. No other characters accepted

### DIFF
--- a/liota/dccs/iotcc.py
+++ b/liota/dccs/iotcc.py
@@ -247,19 +247,15 @@ class IotControlCenter(DataCenterComponent):
         }
         for key, value in properties.items():
             if not self._validate_input(key):
-                log.error("Property key {0} contains unacceptable character for resource {1}".format(key, entity_name))
-                continue
+                raise ValueError("Property key {0} contains unacceptable character for resource {1}".format(key, entity_name))
             if not len(key) <= 100:
-                log.error("Property key {0} for resource {1} contains more than 100 characters".format(key, entity_name))
-                continue
+                raise ValueError("Property key {0} for resource {1} contains more than 100 characters".format(key, entity_name))
             if not self._validate_input(value):
-                log.error("Property value {0} of key {1} contains unacceptable character for resource {2}"
+                raise ValueError("Property value {0} of key {1} contains unacceptable character for resource {2}"
                           .format(value, key, entity_name))
-                continue
             if not len(value) <= 255:
-                log.error("Property value {0} of key {1} for resource {2} contains more than 255 characters"
+                raise ValueError("Property value {0} of key {1} for resource {2} contains more than 255 characters"
                           .format(value, key, entity_name))
-                continue
             msg["body"]["property_data"].append({"propertyKey": key, "propertyValue": value})
         return msg
 
--- a/liota/dccs/iotcc.py
+++ b/liota/dccs/iotcc.py
@@ -91,9 +91,13 @@ class IotControlCenter(DataCenterComponent):
 
         """
         if not self._validate_input(entity_obj.name):
-            raise Exception("Name of the resource contains unacceptable character : " + entity_obj.name)
+            raise ValueError("Name of the resource contains unacceptable character : " + entity_obj.name)
+        if not len(entity_obj.name) <= 50:
+            raise ValueError("Name of the resource contains more than 50 characters : " + entity_obj.name)
         if not self._validate_input(entity_obj.entity_type):
-            raise Exception("Type of the resource contains unacceptable character : " + entity_obj.entity_type)
+            raise ValueError("Type of the resource contains unacceptable character : " + entity_obj.entity_type)
+        if not len(entity_obj.entity_type) <= 50:
+            raise ValueError("Type of the resource contains more than 50 characters : " + entity_obj.entity_type)
 
         if isinstance(entity_obj, Metric):
             # reg_entity_id should be parent's one: not known here yet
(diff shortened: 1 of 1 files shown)


liota/dccs/iotcc.py
```diff
@@ -89,6 +89,9 @@ def register(self, entity_obj):
         """ Register the objects
 
         """
+        if not self._validate_input(entity_obj.name):
+            raise Exception("Name of the resource contains unacceptable character : " + entity_obj.name)
```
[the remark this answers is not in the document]
arunv10: Agreed and I have added validation of entity_obj.entity_type

liota/dccs/iotcc.py
```diff
@@ -672,3 +681,10 @@ def on_response(msg):
         self.comms.send(json.dumps(self._get_properties(self.next_id(), resource_uuid)))
         on_response(self.recv_msg_queue.get(True, timeout))
         return self.prop_list
+
+    def _validate_input(self, input):
+        """ validates if the input string contains only the whitelisted characters """
+        import re
```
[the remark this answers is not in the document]
arunv10: I have changed the import to the preferred way. Thanks for sharing the links.

GregBollella:
Looks fine. However, we should consider if v2 will accept a superset of these characters and determine how to accommodate.

liota/dccs/iotcc.py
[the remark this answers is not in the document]
KohliDev: I believe we should raise ValueError

https://docs.python.org/2/library/exceptions.html#exceptions.ValueError

Also, can you include the check on the length of the name or the property as I believe there is the restriction from the server side?

liota/dccs/iotcc.py
```diff
@@ -236,6 +242,12 @@ def _properties(self, msg_id, entity_type, entity_id, entity_name, timestamp, pr
             }
         }
         for key, value in properties.items():
+            if not self._validate_input(key):
+                log.error("Property key {0} contains unacceptable character for resource {1}".format(key, entity_name))
+                continue
```
KohliDev: What is the reason for not raising ValueError here?

liota/dccs/iotcc.py
```diff
@@ -672,3 +684,9 @@ def on_response(msg):
         self.comms.send(json.dumps(self._get_properties(self.next_id(), resource_uuid)))
         on_response(self.recv_msg_queue.get(True, timeout))
         return self.prop_list
+
+    def _validate_input(self, input):
```
KohliDev: If it contains the following "^[A-Za-z0-9\s\._-]+$" characters can the command injection happen in python or only in the shell script scenario?

liota/dccs/iotcc.py
arunv10: Changed Exception to ValueError

liota/dccs/iotcc.py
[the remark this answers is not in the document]
arunv10: Updated PR with the ValueError and length of the name/property as per the server-side limits.

liota/dccs/iotcc.py
arunv10: To avoid various possibilities of injections, it was suggested to allow only these white-listed set of characters to be consumed from the user code.

liota/dccs/iotcc.py
arunv10: @GregBollella @KohliDev @pmasrani 
I have updated the PR with limit checks for the number of characters for resource names and properties.
These limits are from airwatch and relevant for v1.
Could you please confirm if this is acceptable.

KohliDev:
@GregBollella Agreed we might need to accommodate special characters in v2. 

Also, the AW DB has the restriction on the entity_name & entity_type as 50 characters, property key_length as 100 and property_value as 255 these changes have also been applied as part of this Pull request please review.

We can remove these restrictions in v2.

GregBollella:
Agree with these limits

liota/dccs/iotcc.py
KohliDev: @arunv10 You are logging the error not raising it is this the expected behavior? I think if it is an important property which user want to set but due to the restriction, it fails we should raise the exception to user code.

arunv10: @KohliDev I think it would be better to do the best effort approach for setting properties as there are optional. So that other valid properties are set. What do you think?

KohliDev: Okay got it. Then basically, we shouldn't set the property if they don't match the criteria and log the error. 

But the current code changes in this PR will still send the property, I think we shouldn't as we know the limit exist on the AW (v1). Also `continue` statement shouldn't be required for logging the error in `if` Loop.

arunv10: In the code changes, if a property's name or value doesn't satisfy the criteria then it is skipped and the next property in the for loop is picked using the continue statement.

If a property passes all tests then it is appended to the property data of the add_properties_request message.

pmasrani: Good catch Vaibhav!

Arun, i think we should throw the error in this case as well, instead of best effort. If we just log the error, the property will nto get set, but the user call to set properties will succeed. He will have no way to know if the property got set or not, but to go through the logs. If we throw the error, the loading of the package which is setting the property will fail, and he will be forced to take a look at it.

arunv10: @KohliDev and @pmasrani  I understood your point. I have addressed them.

liota/dccs/iotcc.py
```diff
@@ -89,6 +90,15 @@ def register(self, entity_obj):
         """ Register the objects
 
         """
+        if not self._validate_input(entity_obj.name):
```
pmasrani: I think we can now modify the function _validate_input(input, length), check the matching string, check the size and throw a common ValueError, size mismatch or invalid character used in input "". The caller can just make a single call like _validate_input(entity_obj.name, 50).

Currently, the code seems to have a lot of duplication, just to print exactly where the issue is happening.

arunv10: @pmasrani  I have addressed the suggested changes. Please review the updated PR.

liota/dccs/iotcc.py
```diff
@@ -672,3 +678,11 @@ def on_response(msg):
...
         on_response(self.recv_msg_queue.get(True, timeout))
         return self.prop_list
+
+    def _assert_input(self, input, max_length):
+        """ validates if the input string contains only the whitelisted characters """
+        if not match('^[A-Za-z0-9\s\._-]+$', input):
+            raise ValueError("The provided string contains unacceptable character : {0}".format(input))
+        if not len(input) <= max_length:
+            raise ValueError("The provided string contains more than {0} characters : {1}" .format(max_length, input))
+        return
```
pmasrani: Nit: I don't think we need a return statement, but can be addressed later.

arunv10:
@KohliDev @pmasrani  If changes are okay, can you please merge this PR?

PR #186, opened 2017-11-15
Add liota V2 features:

winniex1:
1. add a DCC for VMware IoTC V2 Data Center Component, which has
   blocking call for Managed Object register, set properties, and
   create relationship operations (multithread handling support)
2. modify dcc.py to add Exception class for set properties and
   create relationship
3. add example liota packages for IoTC V2 under "packages/
   user_packages/iotcv2" folder

### DIFF
--- a/liota/dccs/iotcv2.py
+++ b/liota/dccs/iotcv2.py
@@ -34,17 +34,11 @@ import json
 import logging
 import time
 import threading
-import ConfigParser
-import os
 import Queue
-from time import gmtime, strftime
 from threading import Lock
 
 from liota.dccs.dcc import DataCenterComponent, RegistrationFailure, SetPropertiesFailure, CreateRelationshipFailure
-from liota.lib.protocols.helix_protocol import HelixProtocol
 from liota.entities.metrics.metric import Metric
-from liota.lib.utilities.utility import LiotaConfigPath, getUTCmillis, mkdir, read_liota_config
-from liota.lib.utilities.si_unit import parse_unit
 from liota.entities.metrics.registered_metric import RegisteredMetric
 from liota.entities.registered_entity import RegisteredEntity
 
@@ -85,13 +79,13 @@ class Iotcv2(DataCenterComponent):
         self.req_dict = {}
         thread = threading.Thread(target=self._dispatch_recvd_msg)
         thread.daemon = True
-        # This thread will continuously run in background to check received responses
+        # This thread will continuously run in background to check and dispatch received responses
         thread.start()
 
     def register(self, entity_obj):
         if isinstance(entity_obj, Metric):
             # reg_entity_id should be parent's one: not known here yet
-            # will add in creat_relationship();
+            # will add in create_relationship();
             log.info("Registering metric {0} locally".format(entity_obj.name))
             return RegisteredMetric(entity_obj, self, None)
         else:
--- a/liota/dccs/iotcv2.py
+++ b/liota/dccs/iotcv2.py
@@ -141,8 +141,6 @@ class Iotcv2(DataCenterComponent):
         if isinstance(reg_entity_child, RegisteredMetric):
             # should save parent's reg_entity_id
             reg_entity_child.reg_entity_id = reg_entity_parent.reg_entity_id
-            #entity_obj = reg_entity_child.ref_entity
-            #self.publish_unit(reg_entity_child, entity_obj.name, entity_obj.unit)
             return
         else:
             log.info("Registering parent and child relationship with IoTCV2 for {0} {1}".\
--- a/liota/dccs/iotcv2.py
+++ b/liota/dccs/iotcv2.py
@@ -1,0 +1,346 @@
+# -*- coding: utf-8 -*-
+# ----------------------------------------------------------------------------#
+#  Copyright © 2015-2016 VMware, Inc. All Rights Reserved.                    #
+#                                                                             #
+#  Licensed under the BSD 2-Clause License (the “License”); you may not use   #
+#  this file except in compliance with the License.                           #
+#                                                                             #
+#  The BSD 2-Clause License                                                   #
+#                                                                             #
+#  Redistribution and use in source and binary forms, with or without         #
+#  modification, are permitted provided that the following conditions are met:#
+#                                                                             #
+#  - Redistributions of source code must retain the above copyright notice,   #
+#      this list of conditions and the following disclaimer.                  #
+#                                                                             #
+#  - Redistributions in binary form must reproduce the above copyright        #
+#      notice, this list of conditions and the following disclaimer in the    #
+#      documentation and/or other materials provided with the distribution.   #
+#                                                                             #
+#  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"#
+#  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE  #
+#  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE #
+#  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE  #
+#  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR        #
+#  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF       #
+#  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS   #
+#  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN    #
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)    #
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF     #
+#  THE POSSIBILITY OF SUCH DAMAGE.                                            #
+# ----------------------------------------------------------------------------#
+
+import json
+import logging
+import time
+import threading
+import ConfigParser
+import os
+import Queue
+from time import gmtime, strftime
+from threading import Lock
+
+from liota.dccs.dcc import DataCenterComponent, RegistrationFailure, SetPropertiesFailure, CreateRelationshipFailure
+from liota.lib.protocols.helix_protocol import HelixProtocol
+from liota.entities.metrics.metric import Metric
+from liota.lib.utilities.utility import LiotaConfigPath, getUTCmillis, mkdir, read_liota_config
+from liota.lib.utilities.si_unit import parse_unit
+from liota.entities.metrics.registered_metric import RegisteredMetric
+from liota.entities.registered_entity import RegisteredEntity
+
+log = logging.getLogger(__name__)
+
+timeout = 300
+
+class Request:
+    def __init__(self, guid, operation_id, user_queue):
+        self.guid = guid
+        # assume operation_id will be unique in one process
+        self.operation_id = operation_id
+        self.user_queue = user_queue
+
+class Iotcv2(DataCenterComponent):
+    """ The implementation of VMware IoTC V2 Data Center Component
+
+    """
+
+    def __init__(self, con):
+        log.info("Logging into IOTC V2 DCC")
+        self.comms = con
+        if not self.comms.identity.username:
+            log.error("Username not found")
+            raise ValueError("Username not found")
+        elif not self.comms.identity.password:
+            log.error("Password not found")
+            raise ValueError("Password not found")
+        thread = threading.Thread(target=self.comms.receive)
+        thread.daemon = True
+        # This thread will continuously run in background to receive response or actions from DCC
+        thread.start()
+        time.sleep(0.5)
+        self.counter = 0
+
+        self.recv_msg_queue = self.comms.userdata
+        self.req_ops_lock = Lock()
+        self.req_dict = {}
+        thread = threading.Thread(target=self._dispatch_recvd_msg)
+        thread.daemon = True
+        # This thread will continuously run in background to check received responses
+        thread.start()
+
+    def register(self, entity_obj):
+        if isinstance(entity_obj, Metric):
+            # reg_entity_id should be parent's one: not known here yet
+            # will add in creat_relationship();
+            log.info("Registering metric {0} locally".format(entity_obj.name))
+            return RegisteredMetric(entity_obj, self, None)
+        else:
+            log.info("Registering resource with IoTCV2 {0}".format(entity_obj.name))
+            # register fun internal queue
+            reg_resp_q = Queue.Queue()
+
+            def on_response(msg, reg_resp_q):
+                try:
+                    log.debug("Received msg: {0}".format(msg))
+                    json_msg = json.loads(msg)
+                    log.debug("Processing msg: {0}".format(json_msg["type"]))
+                    if json_msg["type"] == "registerMO_rsp" and json_msg["guid"] != "null" and \
+                                    json_msg["guid"] == entity_obj.entity_id:
+                        log.info("Received registration response for: {0}".format(json_msg["guid"]))
+                        if json_msg["result"] == 'SUCCESS':
+                            self.reg_entity_id = json_msg["guid"]
+                    else:
+                        log.info("Waiting for resource registration response")
+                        on_response(reg_resp_q.get(True, timeout), reg_resp_q)
+                except:
+                    raise Exception("Exception while registering resource")
+
+            # create and add register request into req_list (waiting list)
+            operation_id = self.next_id()
+            req = Request(entity_obj.entity_id, operation_id, reg_resp_q)
+            log.info("operation_id:{0} req:{1}".format(operation_id, req))
+            with self.req_ops_lock:
+                self.req_dict.update({str(operation_id): req})
+                log.info("operation_id:{0} req_dict:{1}".format(operation_id, self.req_dict))
+            self.comms.send(json.dumps(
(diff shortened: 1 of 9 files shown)


liota/dccs/iotcv2.py
```diff
@@ -0,0 +1,346 @@
...
+    def __init__(self, con):
+        log.info("Logging into IOTC V2 DCC")
+        self.comms = con
+        if not self.comms.identity.username:
+            log.error("Username not found")
+            raise ValueError("Username not found")
+        elif not self.comms.identity.password:
+            log.error("Password not found")
+            raise ValueError("Password not found")
+        thread = threading.Thread(target=self.comms.receive)
```
pmasrani: Why are we starting two thread's here? One on line 79 and another on line 89? We only need one thread for receiving the messages, right?

liota/dccs/iotcv2.py
```diff
@@ -0,0 +1,346 @@
...
+        thread.daemon = True
+        # This thread will continuously run in background to receive response or actions from DCC
+        thread.start()
+        time.sleep(0.5)
+        self.counter = 0
+
+        self.recv_msg_queue = self.comms.userdata
+        self.req_ops_lock = Lock()
+        self.req_dict = {}
+        thread = threading.Thread(target=self._dispatch_recvd_msg)
```
pmasrani: Can we please name this thread better? Something like message_receive_thread?

liota/dccs/iotcv2.py
```diff
@@ -0,0 +1,346 @@
...
+        thread.daemon = True
+        # This thread will continuously run in background to check received responses
+        thread.start()
+
+    def register(self, entity_obj):
+        if isinstance(entity_obj, Metric):
+            # reg_entity_id should be parent's one: not known here yet
+            # will add in creat_relationship();
```
pmasrani: spell check; creat_relationship

liota/dccs/iotcv2.py
```diff
@@ -0,0 +1,346 @@
...
+                    if json_msg["type"] == "registerMO_rsp" and json_msg["guid"] != "null" and \
+                                    json_msg["guid"] == entity_obj.entity_id:
+                        log.info("Received registration response for: {0}".format(json_msg["guid"]))
+                        if json_msg["result"] == 'SUCCESS':
+                            self.reg_entity_id = json_msg["guid"]
+                    else:
+                        log.info("Waiting for resource registration response")
+                        on_response(reg_resp_q.get(True, timeout), reg_resp_q)
+                except:
+                    raise Exception("Exception while registering resource")
```
pmasrani: Why are we catching the actual exception and raising it as a new Exception? Will the stacktrace not get lost here?

liota/dccs/iotcv2.py
```diff
@@ -0,0 +1,346 @@
...
+            with self.req_ops_lock:
+                self.req_dict.update({str(operation_id): req})
+                log.info("operation_id:{0} req_dict:{1}".format(operation_id, self.req_dict))
+            self.comms.send(json.dumps(
+                self._registration(operation_id, entity_obj.entity_id, entity_obj.name, entity_obj.entity_type)))
+            # block until there is an item available or timeout
+            try:
+                on_response(reg_resp_q.get(True, timeout), reg_resp_q)
+            except:
+                raise Exception("Exception while registering resource")
```
pmasrani: on_response() catches an exception and then throws another one, and that is caught here and we throw another one again? Why can't we just let the original exception get logged?

liota/dccs/iotcv2.py
```diff
@@ -0,0 +1,346 @@
...
+            def on_response(msg, reg_resp_q):
+                try:
+                    log.debug("Received msg: {0}".format(msg))
+                    json_msg = json.loads(msg)
+                    log.debug("Processing msg: {0}".format(json_msg["type"]))
+                    if json_msg["type"] == "registerMO_rsp" and json_msg["guid"] != "null" and \
+                                    json_msg["guid"] == entity_obj.entity_id:
+                        log.info("Received registration response for: {0}".format(json_msg["guid"]))
+                        if json_msg["result"] == 'SUCCESS':
+                            self.reg_entity_id = json_msg["guid"]
```
pmasrani: What happens when it is not SUCCESS?

liota/dccs/iotcv2.py
```diff
@@ -0,0 +1,346 @@
...
+            if not self.reg_entity_id:
+                raise RegistrationFailure()
+            log.info("Resource Registered {0}".format(entity_obj.name))
+
+            return RegisteredEntity(entity_obj, self, self.reg_entity_id)
+
+    def create_relationship(self, reg_entity_parent, reg_entity_child):
+        # sanity check: must be RegisteredEntity or RegisteredMetricRegisteredMetric
+        if (not isinstance(reg_entity_parent, RegisteredEntity) \
+                    and not isinstance(reg_entity_parent, RegisteredMetric)) \
```
pmasrani: Can reg_entity_parent be RegisteredMetric?

liota/dccs/iotcv2.py
```diff
@@ -0,0 +1,346 @@
...
+                or (not isinstance(reg_entity_child, RegisteredEntity) \
+                            and not isinstance(reg_entity_child, RegisteredMetric)):
+            raise TypeError()
+
+        reg_entity_child.parent = reg_entity_parent
+        if isinstance(reg_entity_child, RegisteredMetric):
+            # should save parent's reg_entity_id
+            reg_entity_child.reg_entity_id = reg_entity_parent.reg_entity_id
+            #entity_obj = reg_entity_child.ref_entity
```
pmasrani: Please remove commented code.

liota/dccs/iotcv2.py
```diff
@@ -0,0 +1,346 @@
...
+                    if json_msg["type"] == "createRelationship_rsp" and json_msg["guid"] != "null" \
+                        and json_msg["guid"] == reg_entity_parent.reg_entity_id:
+                        log.info("Received createRelationship response")
+                        if json_msg["result"] == 'SUCCESS':
+                            self.set_relationship = True
+                    else:
+                        log.info("Waiting for create relationship response")
+                        on_response(rel_resp_q.get(True, timeout), rel_resp_q)
+                except:
+                    raise Exception("Exception while create relationship")
```
pmasrani: same as above.

liota/dccs/iotcv2.py
```diff
@@ -0,0 +1,346 @@
...
+                raise CreateRelationshipFailure()
+            log.info("Relationship Created")
+
+    def _format_data(self, reg_metric):
+        met_cnt = reg_metric.values.qsize()
+        if met_cnt == 0:
+            return
+        data = []
+        for _ in range(met_cnt):
+            m = reg_metric.values.get(block=True)
```
pmasrani: Do we need to specify block=True? Isn't it True by default?

liota/dccs/iotcv2.py
```diff
@@ -0,0 +1,346 @@
...
+            entity = reg_entity.parent.ref_entity
+        else:
+            entity = reg_entity.ref_entity
+
+        log.info("Properties defined for resource {0}".format(entity.name))
+        # set_properties fun internal queue
+        set_prop_resp_q = Queue.Queue()
+        self.set_prop = False
+
+        def on_response(msg, set_prop_resp_q):
```
pmasrani: If we are doing the same kind of check everywhere, why can't we extract this out as a single function and call it with parameters everywhere? Why are we duplicating the code? Am i missing something here?

liota/dccs/iotcv2.py
```diff
@@ -0,0 +1,346 @@
...
+                # assume operation_id will be unique in one process
+                log.debug("self.req_dict {0} keys:{1}".format(self.req_dict, self.req_dict.keys()))
+                if (json_msg["operation_id"] in self.req_dict.keys()):
+                    req = self.req_dict[json_msg["operation_id"]]
+                    log.debug("self.req_dict {0} req:{1}".format(self.req_dict, req))
+                    # get/delete request from dictionary
+                    if (req is not None):
+                        log.debug("json_msg['guid'] {0} req.guid:{1}".format(json_msg["guid"], req.guid))
+                        # double confirm guid
+                        if (json_msg["guid"] != req.guid):
```
pmasrani: why are we confirming guid here, and then again inside on_response?

packages/user_packages/iotcv2/iotcv2_toyota_sim.py
```diff
@@ -0,0 +1,98 @@
...
+                                                   identity=identity,
+                                                   tls_conf=tls_conf,
+                                                   enable_authentication=True
+                                                   )
+                                      )
+        try:
+            self.iotcc_edge_system = self.iotcc.register(edge_system)
+            registry.register("iotc_v2_mqtt_wssl", self.iotcc)
+            registry.register("iotc_v2_mqtt_wssl_edge_system", self.iotcc_edge_system)
+        except RegistrationFailure:
```
pmasrani: Here, we are just catching RegistrationFailure, so in case of other Exception thrown, it will just go uncaught.

packages/user_packages/iotcv2/iotcv2_toyota_sim.py
```diff
@@ -0,0 +1,98 @@
...
+            print "EdgeSystem registration to IOTCV2 failed"
```
pmasrani: Should we have print statements in package?

packages/user_packages/iotcv2/iotcv2_toyota_sim.py
```diff
@@ -0,0 +1,98 @@
...
+            return
+        self.iotcc.set_properties(self.iotcc_edge_system, self.config['EdgeSystemPropList'])
```
pmasrani: We are not catching the Exceptions in set_properties.

packages/user_packages/iotcv2/sampleProp.yml
```diff
@@ -0,0 +1,12 @@
+EdgeSystemName : "EdgeSystem-Name"
```
pmasrani: Any reason why we added sampleProp.yml and not reuse sampleProp.conf?

liota/dccs/iotcv2.py
```diff
@@ -0,0 +1,346 @@
...
+import time
+import threading
+import ConfigParser
+import os
+import Queue
+from time import gmtime, strftime
+from threading import Lock
+
+from liota.dccs.dcc import DataCenterComponent, RegistrationFailure, SetPropertiesFailure, CreateRelationshipFailure
+from liota.lib.protocols.helix_protocol import HelixProtocol
```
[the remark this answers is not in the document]
winniex1: removed all unused import

liota/dccs/iotcv2.py
winniex1: Now we must have 2 threads, which can allow us to support multiple register calls in parallel without losing any responses. Each register call will have a internal queue to block and wait on for responses.
The 2 threads are:
1 is receiving message thread on Line 79 to put received messages into recv_msg_queue; 
2 is to check received messages in recv_msg_queue and dispatch them into each call's response queue.

liota/dccs/iotcv2.py
winniex1: No, it's not message_receive_thread, but a internal thread for dispatching received messages/responses into each blocking call's internal waiting queue.

liota/dccs/iotcv2.py
winniex1: done

liota/dccs/iotcv2.py
winniex1: self.reg_entity_id will not be set, then RegistrationFailure() will be raised.

liota/dccs/iotcv2.py
winniex1: I treat this like what we have done for IotCC. I believe we decided to let user code capture this exception and decide what to do with it. So did not use log.exception to print stacktrace.

liota/dccs/iotcv2.py
winniex1: same as above

packages/user_packages/iotcv2/sampleProp.yml
winniex1: I have a example code for toyota simulation where we have some specific info, e.g., for properties list. In addition, this is just examples codes, and wanted to expose to users more options for configurations.

packages/user_packages/iotcv2/iotcv2_toyota_sim.py
winniex1: this is what we have done for iotcc example codes

liota/dccs/iotcv2.py
winniex1: isn't better to set it explicitly like what we have done for IoTCC?

liota/dccs/iotcv2.py
winniex1: They are not the same, but have similar procedure. They have different if condition check. Do not want to mess them up with too many input arguments. The on_response codes only have ~ 10 lines.

liota/dccs/iotcv2.py
winniex1: as comment said, double confirm to guarantee between there is not disruption or destroy.

packages/user_packages/iotcv2/iotcv2_toyota_sim.py
winniex1: no.. Exception inside on_response will be inside register(). We will only finally raise RegistrationFailure to user call.

liota/dccs/iotcv2.py
winniex1: so far, in v2, No.

liota/dccs/iotcv2.py
winniex1: ok

liota/dccs/iotcv2.py
winniex1: I thought this over, and use log.exception to print stacktrace.

liota/dccs/iotcv2.py
winniex1: I thought this over, and just use log.exception to get logged.

liota/dccs/iotcv2.py
winniex1: changed

packages/user_packages/iotcv2/iotcv2_toyota_sim.py
GregBollella: Should we catch an exception for this set_properties call?

packages/user_packages/iotcv2/iotcv2_toyota_sim.py
```diff
@@ -0,0 +1,98 @@
...
+                log.debug("ecu_sn:{0} PropList:{1}".format(ecu_sn, self.config['ECUPropList']))
+                prop_dict = {}
+                prop_dict.update({"Make": ecu_makes[i % 3]})
+                prop_dict.update({"Model": ecu_models[i % 3][i % 5]})
+                prop_dict.update({"SerialNumber": ecu_sn})
+                log.debug("new PropList:{0}".format(prop_dict))
+                self.iotcc.set_properties(self.reg_ecu, prop_dict)
+                self.iotcc.create_relationship(self.iotcc_edge_system, self.reg_ecu)
+        except Exception:
+            log.exception("Connected Device Registration and Metrics are Not Available")
```
GregBollella: Log message seems incorrect because some of the 19 might be registered.

packages/user_packages/iotcv2/iotcv2_wssl.py
```diff
@@ -0,0 +1,65 @@
...
+                                                   port=self.config['BrokerPort'],
+                                                   identity=identity,
+                                                   tls_conf=tls_conf,
+                                                   enable_authentication=True
+                                                   )
+                                      )
+        try:
+            self.iotcc_edge_system = self.iotcc.register(edge_system)
+            registry.register("iotc_v2_mqtt_wssl", self.iotcc)
+            registry.register("iotc_v2_mqtt_wssl_edge_system", self.iotcc_edge_system)
```
GregBollella: Should there be a set properties call here somewhere? Or is it elsewhere?

packages/user_packages/iotcv2/iotcv2_toyota_sim.py
winniex1: already added.

packages/user_packages/iotcv2/iotcv2_toyota_sim.py
winniex1: modified

packages/user_packages/iotcv2/iotcv2_wssl.py
winniex1: added

packages/user_packages/iotcv2/iotcv2_edge_system.py
```diff
@@ -0,0 +1,25 @@
+import yaml
```
KohliDev: Lacks License

packages/user_packages/iotcv2/iotcv2_toyota_sim.py
```diff
@@ -0,0 +1,102 @@
+from liota.dccs.iotcv2 import Iotcv2
```
KohliDev: Lacks License

packages/user_packages/iotcv2/iotcv2_wssl.py
```diff
@@ -0,0 +1,68 @@
+import yaml
```
KohliDev: Lacks License

packages/user_packages/iotcv2/sampleProp.yml
KohliDev: This should be moved to conf file keeping it consistent across the codebase.

liota/dccs/iotcv2.py
```diff
@@ -0,0 +1,333 @@
...
+        self.user_queue = user_queue
+
+class Iotcv2(DataCenterComponent):
+    """ The implementation of VMware IoTC V2 Data Center Component
+
+    """
+
+    def __init__(self, con):
+        log.info("Logging into IOTC V2 DCC")
+        self.comms = con
```
KohliDev: This codebase should be merged with recent refactoring of v1 as it lacks 73 commits behind.

PR #188, opened 2017-11-17
Modified all calls to use only liota uuid

pmasrani:


liota/dccs/iotcc.py
```diff
@@ -177,9 +177,8 @@ def create_relationship(self, reg_entity_parent, reg_entity_child):
             Parameters:
             - obj: The object that has just obtained an UUID
         """
-        # sanity check: must be RegisteredEntity or RegisteredMetricRegisteredMetric
-        if (not isinstance(reg_entity_parent, RegisteredEntity) \
-                    and not isinstance(reg_entity_parent, RegisteredMetric)) \
+        # sanity check: must be RegisteredEntity or RegisteredMetric
+        if (not isinstance(reg_entity_parent, RegisteredEntity)) \
```
gkmccready: This doesn't look like it'll parse correctly? if (()) will close the statement and should be a syntax error on the next line?

liota/dccs/iotcc.py
```diff
@@ -59,7 +59,7 @@ class IotControlCenter(DataCenterComponent):
 
     def __init__(self, con):
         log.info("Logging into DCC")
-        self._version = 20171023
+        self._version = 20171118
```
KohliDev: We`ll have to change the version mechanism (probably later) keeping dates as version number is confusing as this date signifies the code review raised but the code will get checked-in lot later.

liota/dccs/iotcc.py
```diff
@@ -211,14 +210,22 @@ def _registration(self, msg_id, res_id, res_name, res_kind):
...
+    def _relationship(self, msg_id, parent_entity, child_entity):
         return {
             "transactionID": msg_id,
             "version": self._version,
             "type": "create_relationship_request",
             "body": {
-                "parent": parent_res_uuid,
-                "child": child_res_uuid
+                "parent":  {
+                    "kind": parent_entity.entity_type,
```
KohliDev: Please provide comments on why the kind, id and name is required to be sent for _relationship call.

liota/dccs/iotcc.py
```diff
@@ -239,13 +246,15 @@ def _properties(self, msg_id, entity_type, entity_id, entity_name, timestamp, pr
...
 
-    def _get_properties(self, msg_id, res_uuid):
+    def _get_properties(self, msg_id, ref_entity):
         return {
             "transactionID": msg_id,
             "version": self._version,
             "type": "get_properties_request",
             "body": {
-                "uuid": res_uuid
+                "kind": ref_entity.entity_type,
```
KohliDev: Please provide the reason for _get_properties requiring "kind" and "name"?

liota/dccs/iotcc.py
```diff
@@ -177,9 +177,8 @@ def create_relationship(self, reg_entity_parent, reg_entity_child):
             Parameters:
             - obj: The object that has just obtained an UUID
         """
-        # sanity check: must be RegisteredEntity or RegisteredMetricRegisteredMetric
-        if (not isinstance(reg_entity_parent, RegisteredEntity) \
-                    and not isinstance(reg_entity_parent, RegisteredMetric)) \
```
KohliDev: What is the reason for removing the metric check or is it by mistakenly removed?

liota/dccs/iotcc.py
pmasrani: In python, if statement is parsed till a matching colon (:) is received.

liota/dccs/iotcc.py
pmasrani: reg_entity_parent can't be a RegisteredMetric.

liota/dccs/iotcc.py
pmasrani: As described in the other CL, all the calls have now been standardized to use the same format (uuid, name, kind)

liota/dccs/iotcc.py
pmasrani: As described in the other CL, all the calls have now been standardized to use the same format (uuid, name, kind)

liota/dccs/iotcc.py
pmasrani: We can discuss this offline in future for future CL.

liota/dccs/iotcc.py
KohliDev: The check should exist for metric as well because the inputs are passed from the user code and we don't have any control on it.

pmasrani: We can not allow a user to have a RegisteredMetric as a parent for any other RegisteredMetric or RegisteredEntity. Hence, this check is removed so that even if a user passes RegisteredMetric as a parent, we can throw an error. child object can still be RegisteredMetric or RegisteredEntity, and that check is continued.

KohliDev: Got it the extra brackets confused me in the CL, it should directly be;

not isinstance(reg_entity_parent, RegisteredEntity)

pmasrani: Yes, you are right. The extra bracket just helps readability, that the not is applicable to the first condition only. Technically, it can be removed, but will keep it for now.

liota/dccs/iotcc.py
gkmccready: Thank you! I missed that when I tried to check the python docs.

liota/dccs/iotcc.py
```diff
@@ -232,14 +231,22 @@ def _registration(self, msg_id, res_id, res_name, res_kind):
...
+                "parent":  {
+                    "kind": parent_entity.entity_type,
+                    "id": parent_entity.entity_id,
+                    "name": parent_entity.name
+                },
+                "child":  {
+                    "kind": child_entity.entity_type,
+                    "id": child_entity.entity_id,
+                    "name": child_entity.name
+                }
```
KohliDev: The other methods (_registration, _ properties) should be refactored to use only the entity object rather than passing all the parameters separately (res_id, res_name, res_kind).

liota/dccs/iotcc.py
KohliDev: Which CL?

pmasrani: Corresponding CL for adapter code changes.

liota/dccs/iotcc.py
pmasrani: Yes, very good point. Avoiding last minute refactoring of existing code. We should definitely take this up in near future.

PR #191, opened 2017-11-23
Let package manager return command execution status:

winniex1:
1. package manager creates one more namedPipe to pass status
2. once user passes commands to messenger namedPipe through liotapkg,
liotapkg will read from reponser namedPipe to get status
3. for commands of ["unload", "delete", "list", "stat", "unload_all",
   "unload_all", "terminate"], 'Success' means that package manager
   has successfully received these commands;
   for commands of ["load", "reload", "update"], 'Success' means that
   verification of the package checksum succeeds, while 'Failure'
   means that verification fails;
   'Unsupported' means that receives some unsupported commands.

### DIFF
--- a/config/liota.conf
+++ b/config/liota.conf
@@ -20,6 +20,7 @@ collect_thread_pool_size = 30
 [PKG_CFG]
 pkg_path = /usr/lib/liota/packages
 pkg_msg_pipe = /usr/lib/liota/packages/liotad/package_messenger.fifo
+pkg_rsp_pipe = /usr/lib/liota/packages/liotad/package_responser.fifo
 pkg_list = /usr/lib/liota/packages/liotad/packages_auto.txt
 
 [DISC_CFG]
--- a/config/liota.conf
+++ b/config/liota.conf
@@ -20,7 +20,7 @@ collect_thread_pool_size = 30
 [PKG_CFG]
 pkg_path = /usr/lib/liota/packages
 pkg_msg_pipe = /usr/lib/liota/packages/liotad/package_messenger.fifo
-pkg_rsp_pipe = /usr/lib/liota/packages/liotad/package_responser.fifo
+pkg_rsp_pipe = /usr/lib/liota/packages/liotad/package_response.fifo
 pkg_list = /usr/lib/liota/packages/liotad/packages_auto.txt
 
 [DISC_CFG]
--- a/liota/core/package_manager.py
+++ b/liota/core/package_manager.py
@@ -494,8 +494,8 @@ class PackageThread(Thread):
         """
         Write command status back to response named pipe.
         Currently, for commands of ["unload", "delete", "list", "stat",
-        "unload_all", "unload_all", "terminate"], 'Success' means
-        that PackageThread has successfully received these commands;
+        "unload_all", "terminate"], 'Success' means that PackageThread
+        has successfully received these commands;
         for commands of ["load", "reload", "update"], 'Success' means
         that verification of the package checksum succeeds, while
         'Failure' means that verification fails; 'Unsupported' means
--- a/liota/core/package_manager.py
+++ b/liota/core/package_manager.py
@@ -73,6 +73,7 @@ package_thread = None
 package_lock = None
 package_path = None
 package_messenger_pipe = None
+package_responser_pipe = None
 
 # Parse Liota configuration file
 package_path = os.path.abspath(
@@ -81,8 +82,12 @@ package_path = os.path.abspath(
 package_messenger_pipe = os.path.abspath(
     read_liota_config('PKG_CFG', 'pkg_msg_pipe')
 )
+package_responser_pipe = os.path.abspath(
+    read_liota_config('PKG_CFG', 'pkg_rsp_pipe')
+)
 assert(isinstance(package_path, basestring))
 assert(isinstance(package_messenger_pipe, basestring))
+assert(isinstance(package_responser_pipe, basestring))
 
 package_startup_list_path = None
 package_startup_list = []
@@ -244,6 +249,7 @@ class PackageThread(Thread):
         self._packages_loaded = {}  # key: package name, value: PackageRecord obj
         self._resource_registry = ResourceRegistry()
         self._resource_registry.register("package_conf", package_path)
+        self._rsp_pipe_file = package_responser_pipe
         self.flag_alive = True
         self.start()
 
@@ -373,10 +379,18 @@ class PackageThread(Thread):
 
             # Switch on message content (command), determine what to do
             command = msg[0]
+            if command in ["unload", "delete", "list", "stat", "unload_all",\
+                           "unload_all", "terminate"]:
+                # currently, write 'Success' to the named pipe once
+                #     successfully receives these commands
+                with open(self._rsp_pipe_file, "w+") as fp:
+                    fp.write('Success\n')
+
+            # currently, write 'Success' to the named pipe after verification
+            #     of the checksum for load/reload/update related commands
             if command in ["load", "reload", "update"]:
                 #-----------------------------------------------------------
                 # Use these commands to handle package management tasks
-
                 with package_lock:
                     offset = 0
                     autoload_flag = False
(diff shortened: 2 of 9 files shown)


config/liota.conf
```diff
@@ -20,6 +20,7 @@ collect_thread_pool_size = 30
 [PKG_CFG]
 pkg_path = /usr/lib/liota/packages
 pkg_msg_pipe = /usr/lib/liota/packages/liotad/package_messenger.fifo
+pkg_rsp_pipe = /usr/lib/liota/packages/liotad/package_responser.fifo
```
gkmccready: Prefer "response.fifo"

liota/core/package_manager.py
```diff
@@ -373,10 +379,18 @@ def run(self):
 
             # Switch on message content (command), determine what to do
             command = msg[0]
+            if command in ["unload", "delete", "list", "stat", "unload_all",\
+                           "unload_all", "terminate"]:
```
gkmccready: "unload_all" is in the list twice.

liota/core/package_manager.py
```diff
@@ -373,10 +379,18 @@ def run(self):
...
+                # currently, write 'Success' to the named pipe once
+                #     successfully receives these commands
+                with open(self._rsp_pipe_file, "w+") as fp:
```
gkmccready: If the open fails we need to do something... log an error?

liota/core/package_manager.py
```diff
@@ -373,10 +379,18 @@ def run(self):
...
+                    fp.write('Success\n')
```
gkmccready: I think we want this to be declared somewhere global since the client and server need to know about the same strings in the response, just like in the requests.

liota/core/package_manager.py
```diff
@@ -472,6 +486,10 @@ def run(self):
                     self.flag_alive = False
                     break
             else:
+                # currently, write 'Unsupported' to the named pipe once
+                #     successfully receives these commands
+                with open(self._rsp_pipe_file, "w+") as fp:
```
gkmccready: What if the open() fails?

liota/core/package_manager.py
```diff
@@ -472,6 +486,10 @@ def run(self):
...
+                    fp.write('Unsupported\n')
```
gkmccready: Same as before; since this is part of a client/server protocol we should have the strings somewhere global for both sides to include and stay consistent.

liota/core/package_manager.py
```diff
@@ -641,9 +659,13 @@ def _package_load(self, file_name, checksum=None, autoload_flag=False,
                         break
                 if (verify_flag == False):
                     log.error("Package %s integrity verification failed" % path_file_ext)
+                    with open(self._rsp_pipe_file, "w+") as fp:
```
gkmccready: Handle open() failure...

liota/core/package_manager.py
```diff
@@ -641,9 +659,13 @@ def _package_load(self, file_name, checksum=None, autoload_flag=False,
...
+                        fp.write('Failure\n')
```
gkmccready: Same comment about putting protocol strings somewhere global for client/server to share.

liota/core/package_manager.py
```diff
@@ -641,9 +659,13 @@ def _package_load(self, file_name, checksum=None, autoload_flag=False,
...
                     return None
             else:
                 sha1 = sha1sum(path_file_ext);
+                with open(self._rsp_pipe_file, "w+") as fp:
+                    fp.write('Success\n')
```
gkmccready: 667/668: Same comment as above. Even more so since we're using the Success string again.

liota/core/package_manager.py
```diff
@@ -1140,13 +1162,20 @@ def _package_delete(self, file_name):
...
         if package_messenger_thread.isAlive():
             with open(package_messenger_pipe, "w") as fp:
                 fp.write(
                     "terminate_messenger_but_you_should_not_do_this_yourself\n")
 
+        log.info("Shutting down package responser...")
+        if package_messenger_thread.isAlive():
+            with open(package_responser_pipe, "w") as fp:
+                fp.write(
+                    "terminate_responser_but_you_should_not_do_this_yourself\n")
```
gkmccready: What does this actually do? Same Q for 1171.

liota/core/package_manager.py
```diff
@@ -1273,6 +1302,35 @@ def initialize():
             return
     assert(stat.S_ISFIFO(os.stat(package_messenger_pipe).st_mode))
 
+    # Validate package responser pipe
```
gkmccready: Please don't replicate 1276-1303 ... make it a function and pass in the path to the fifo you care about.

packages/liotad/liotapkg.sh
```diff
@@ -33,18 +33,19 @@
 
 liota_config="/etc/liota/liota.conf"
 package_messenger_pipe=""
+package_responser_pipe=""
```
gkmccready: Same comment here about "response" versus "responser".

packages/liotad/liotapkg.sh
```diff
@@ -33,18 +33,19 @@
...
 
 if ! [ `ps -ef | grep "liotad.py" | grep -v grep | wc -l` -gt 0 ]
```
gkmccready: Not related to this change, but I really wish we'd move to /var/run/liotad.pid and using "kill -0" rather than ps|grep.

config/liota.conf
winniex1: changed

liota/core/package_manager.py
winniex1: corrected

liota/core/package_manager.py
winniex1: added

liota/core/package_manager.py
winniex1: wrap "writing command status to response named pipe" into _write_cmd_exec_status(), where explained responses meaning. The same content is also added in packages/liotad/README.md file.

liota/core/package_manager.py
winniex1: added try...except, log.exception()

liota/core/package_manager.py
winniex1: done the same as above

liota/core/package_manager.py
winniex1: done the same as above

liota/core/package_manager.py
winniex1: done the same as above

liota/core/package_manager.py
winniex1: done the same as above

liota/core/package_manager.py
winniex1: Thought this over again and feel it is not necessary, so removed it.
The previous purpose is that when execute a "terminate" cmd, notify the user that package thread is executing it. But since we will pass "Success" back, no need to do this.

liota/core/package_manager.py
winniex1: Done. Used a unified function of validate_named_pipe() for all named pipe validation.

packages/liotad/liotapkg.sh
winniex1: modified.

packages/liotad/liotapkg.sh
winniex1: Yes, but should do this in another pull request.

liota/core/package_manager.py
```diff
@@ -472,9 +484,29 @@ def run(self):
...
+                #     successfully receives these commands
+                self._write_cmd_exec_status('Unsupported\n')
                 log.warning("Unsupported command is dropped")
         log.info("Thread exits: %s" % str(self.name))
 
+    def _write_cmd_exec_status(self, msg):
+        """
+        Write command status back to response named pipe.
+        Currently, for commands of ["unload", "delete", "list", "stat",
+        "unload_all", "unload_all", "terminate"], 'Success' means
```
gkmccready: "unload_all" is listed twice.

winniex1: corrected.

winniex1:
Yes, in longer term we plan to use liotapkg.py, then we can apply globals.
BTW, I have corrected and submitted the modification again, please review again.

Thx,
Winnie

PR #195, opened 2017-11-28
Response for properties and relationship calls

KohliDev:
Response check for relationship and properties call

### DIFF
--- a/config/liota.conf
+++ b/config/liota.conf
@@ -5,11 +5,17 @@ json_path = /etc/liota/logging.json
 log_path = /var/log/liota
 
 [IOTCC_PATH]
+# The registered devices details
 dev_file_path = /usr/lib/liota/devs
+# The registered entity details
 entity_file_path = /usr/lib/liota/entity
+# Containing all the meta data about all the registered entities
 iotcc_path = /usr/lib/liota/iotcc.json
+# On reboot fetch the updated properties if any from IoTCC for each entity
 enable_reboot_getprop = False
+# The number of attempts to retry in order to read iotcc.json files
 iotcc_load_retry = 3
+# Configurable queue timeout in order to wait for the response messages from IOTCC
 iotcc_response_timeout = 300
 
 [CRL_PATH]
--- a/config/liota.conf
+++ b/config/liota.conf
@@ -10,6 +10,7 @@ entity_file_path = /usr/lib/liota/entity
 iotcc_path = /usr/lib/liota/iotcc.json
 enable_reboot_getprop = False
 iotcc_load_retry = 3
+iotcc_response_timeout = 300
 
 [CRL_PATH]
 crl_path =
--- a/liota/dccs/iotcc.py
+++ b/liota/dccs/iotcc.py
@@ -177,34 +177,23 @@ class IotControlCenter(DataCenterComponent):
 
         log.info("Unregistering resource with IoTCC {0}".format(entity_obj.ref_entity.name))
         unreg_resp_q = Queue.Queue()
-
-        def on_response(msg, unreg_resp_q):
-            try:
-                log.debug("Received msg: {0}".format(msg))
-                json_msg = json.loads(msg)
-                log.debug("Processing msg: {0}".format(json_msg["type"]))
-                self._check_version(json_msg)
-                if json_msg["type"] == "remove_resource_response" and json_msg["body"]["result"] == "succeeded":
-                    log.info("Unregistration of resource {0} with IoTCC succeeded".format(entity_obj.ref_entity.name))
-                    if entity_obj.ref_entity.entity_type != "HelixGateway":
-                        self._store_device_info(entity_obj.reg_entity_id, entity_obj.ref_entity.name,
-                                                entity_obj.ref_entity.entity_type, None, True)
-                    else:
-                        self._remove_reg_entity_details(entity_obj.ref_entity.name, entity_obj.reg_entity_id)
-                        self._store_device_info(entity_obj.reg_entity_id, entity_obj.ref_entity.name, None, None, True)
-                else:
-                    log.info("Waiting for unregistration response")
-                    on_response(unreg_resp_q.get(True, timeout), unreg_resp_q)
-            except Exception as err:
-                raise err
-
         transaction_id = self._next_id()
         req = Request(transaction_id, unreg_resp_q)
         log.debug("Updating unregister response queue for transaction_id:{0}".format(transaction_id))
         with self.req_ops_lock:
             self.req_dict.update({transaction_id: req})
         self.comms.send(json.dumps(self._unregistration(transaction_id, entity_obj.ref_entity)))
-        on_response(unreg_resp_q.get(True, timeout), unreg_resp_q)
+        response = self._handle_response(unreg_resp_q.get(True, timeout))
+        if response:
+            log.info("Unregistration of resource {0} with IoTCC succeeded".format(entity_obj.ref_entity.name))
+            if entity_obj.ref_entity.entity_type != "HelixGateway":
+                self._store_device_info(entity_obj.reg_entity_id, entity_obj.ref_entity.name,
+                                        entity_obj.ref_entity.entity_type, None, True)
+            else:
+                self._remove_reg_entity_details(entity_obj.ref_entity.name, entity_obj.reg_entity_id)
+                self._store_device_info(entity_obj.reg_entity_id, entity_obj.ref_entity.name, None, None, True)
+        else:
+            raise Exception("Unregistration of resource {0} unsuccessful with IoTCC".format(entity_obj.ref_entity.name))
 
     def create_relationship(self, reg_entity_parent, reg_entity_child):
         """
@@ -241,25 +230,6 @@ class IotControlCenter(DataCenterComponent):
         else:
             # create relationship internal queue
             rel_resp_q = Queue.Queue()
-
-            def on_response(msg, rel_resp_q):
-                try:
-                    log.debug("Received msg: {0}".format(msg))
-                    json_msg = json.loads(msg)
-                    log.debug("Processing msg: {0}".format(json_msg["type"]))
-                    self._check_version(json_msg)
-                    if json_msg["type"] == "create_relationship_response" and json_msg["body"][
-                        "result"] == "succeeded" and json_msg["body"]["parent"] == reg_entity_parent.reg_entity_id and \
-                                    json_msg["body"]["child"] == reg_entity_child.reg_entity_id:
-                        log.info(
-                            "Relationship between entities {0} & {1} created successfully in IoTCC".format(
-                                reg_entity_parent.ref_entity.name, reg_entity_child.ref_entity.name))
-                    else:
-                        log.info("Waiting for create relationship response")
-                        on_response(rel_resp_q.get(True, timeout), rel_resp_q)
-                except Exception as err:
-                    raise err
-
             transaction_id = self._next_id()
             req = Request(transaction_id, rel_resp_q)
             log.debug("Updating create relationship response queue for transaction_id:{0}".format(transaction_id))
@@ -268,9 +238,33 @@ class IotControlCenter(DataCenterComponent):
             self.comms.send(json.dumps(self._relationship(transaction_id,
                                                           reg_entity_parent.reg_entity_id,
                                                           reg_entity_child.reg_entity_id)))
-            on_response(rel_resp_q.get(True, timeout), rel_resp_q)
+            response = self._handle_response(rel_resp_q.get(True, timeout))
+            if response:
+                log.info("Relationship between entities {0} & {1} created successfully in IoTCC".format(
+                    reg_entity_parent.ref_entity.name, reg_entity_child.ref_entity.name))
+            else:
+                raise Exception("Relationship creation between entities {0} & {1} failed in IoTCC".format(
+                    reg_entity_parent.ref_entity.name, reg_entity_child.ref_entity.name))
             if hasattr(reg_entity_parent, 'sys_properties') and reg_entity_parent.sys_properties:
                 self.set_system_properties(reg_entity_child, reg_entity_parent.sys_properties)
+
+    def _handle_response(self, msg):
+        """
+        Process the responses for various types of request messages
+        :param msg: response message received
+        :return: boolean
+        """
+        try:
+            log.debug("Received msg: {0}".format(msg))
+            json_msg = json.loads(msg)
+            log.debug("Processing msg: {0}".format(json_msg["type"]))
+            self._check_version(json_msg)
+            if json_msg["body"]["result"] == "succeeded":
+                return True
+            else:
+                return False
+        except Exception as err:
+            raise err
 
     def _registration(self, msg_id, res_id, res_name, res_kind):
         return {
@@ -368,22 +362,6 @@ class IotControlCenter(DataCenterComponent):
             entity = reg_entity_obj.ref_entity
         # set_properties internal response queue
         set_sys_prop_resp_q = Queue.Queue()
-
-        def on_response(msg, set_sys_prop_resp_q):
-            try:
-                log.debug("Received msg: {0}".format(msg))
-                json_msg = json.loads(msg)
-                log.debug("Processing msg: {0}".format(json_msg["type"]))
-                self._check_version(json_msg)
-                if json_msg["type"] == "add_properties_response" and json_msg["body"][
-                    "result"] == "succeeded":
-                    log.info("System Properties defined for resource {0}".format(entity.name))
-                else:
-                    log.info("Waiting for set system properties response")
-                    on_response(set_sys_prop_resp_q.get(True, timeout), set_sys_prop_resp_q)
-            except Exception as err:
-                raise err
-
         # create and add register request into req_list (waiting list)
         transaction_id = self._next_id()
         req = Request(transaction_id, set_sys_prop_resp_q)
@@ -393,7 +371,12 @@ class IotControlCenter(DataCenterComponent):
         self.comms.send(json.dumps(
             self._properties(transaction_id, entity.entity_type, entity.entity_id, entity.name,
                              getUTCmillis(), system_properties)))
-        on_response(set_sys_prop_resp_q.get(True, timeout), set_sys_prop_resp_q)
+
+        response = self._handle_response(set_sys_prop_resp_q.get(True, timeout))
+        if response:
+            log.info("System Properties defined for resource {0}".format(entity.name))
+        else:
+            raise Exception("Setting System Properties for resource {0} failed".format(entity.name))
 
     def set_properties(self, reg_entity_obj, properties):
         """
@@ -411,22 +394,6 @@ class IotControlCenter(DataCenterComponent):
             entity = reg_entity_obj.ref_entity
 
         set_prop_resp_q = Queue.Queue()
-
-        def on_response(msg, set_prop_resp_q):
-            try:
-                log.debug("Received msg: {0}".format(msg))
-                json_msg = json.loads(msg)
-                log.debug("Processing msg: {0}".format(json_msg["type"]))
-                self._check_version(json_msg)
-                if json_msg["type"] == "add_properties_response" and json_msg["body"][
-                    "result"] == "succeeded":
-                    log.info("Properties defined for resource {0}".format(entity.name))
-                else:
-                    log.info("Waiting for set properties response")
-                    on_response(set_prop_resp_q.get(True, timeout), set_prop_resp_q)
-            except Exception as err:
-                raise err
-
         # create and add register request into req_list (waiting list)
         transaction_id = self._next_id()
         req = Request(transaction_id, set_prop_resp_q)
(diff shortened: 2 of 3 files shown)


liota/dccs/iotcc.py
```diff
@@ -315,10 +340,26 @@ def set_system_properties(self, reg_entity_obj, system_properties):
             reg_entity_obj.sys_properties = system_properties
             entity = reg_entity_obj.ref_entity
 
-        log.info("System Properties defined for resource {0}".format(entity.name))
+        def on_response(msg):
```
GregBollella: Can any of these 'on_response' methods get a msg on the mse_queue that isn't for them? If so, is that message consumed by the particular instances of on_response? If the msg was for another on_response instance is that msg lost? If there are specific instances of the msg_queue for each instance of on_response then why do we even need to test? If the test is required and other msgs can arrive what happens to them?

packages/user_packages/iotcc_mqtt.py
```diff
@@ -67,17 +67,20 @@ def run(self, registry):
...
-                            cert_file=registry.get("edge_system_cert_file"), key_file=registry.get("edge_system_key_file"))
+                            cert_file=registry.get("edge_system_cert_file"),
+                            key_file=registry.get("edge_system_key_file"))
 
         # Encapsulate TLS parameters
         tls_conf = TLSConf(cert_required="CERT_REQUIRED", tls_version="PROTOCOL_TLSv1_2", cipher=None)
 
         # Initialize DCC object with MQTT transport
-        mqtt_msg_attr = MqttMessagingAttributes(pub_topic="liota/"+registry.get("broker_username")+"/request",sub_topic="liota/"+registry.get("broker_username")+"/response")
+        mqtt_msg_attr = MqttMessagingAttributes(pub_topic="liota/" + registry.get("broker_username") + "/request",
```
pmasrani: Shouldn't this be part of liota-core, instead of a user package? I thought we have default topics for iotcc.

liota/dccs/iotcc.py
```diff
@@ -335,10 +376,26 @@ def set_properties(self, reg_entity_obj, properties):
...
             entity = reg_entity_obj.ref_entity
 
-        log.info("Properties defined for resource {0}".format(entity.name))
+        def on_response(msg):
+            try:
+                log.debug("Received msg: {0}".format(msg))
+                json_msg = json.loads(msg)
+                log.debug("Processing msg: {0}".format(json_msg["type"]))
+                self._check_version(json_msg)
+                if json_msg["type"] == "add_properties_response" and json_msg["body"][
```
pmasrani: We should have transaction id in all these calls, and validate based on transaction id. In case of out-of-order processing and repetition of messages, responses can get jumbled.

liota/dccs/iotcc.py
```diff
@@ -335,10 +376,26 @@ def set_properties(self, reg_entity_obj, properties):
         else:
             entity = reg_entity_obj.ref_entity
 
-        log.info("Properties defined for resource {0}".format(entity.name))
+        def on_response(msg):
```
pmasrani: A lot of code seems to be common in on_response. Can't we extract it out as a function, instead of repeating code everywhere? If we have transaction id validation, and "result"=succeeded check, it should be enough.

packages/user_packages/iotcc_mqtt.py
```diff
@@ -97,6 +100,9 @@ def run(self, registry):
                     self.iotcc.set_properties(self.iotcc_edge_system, {"key1": "value1", "key2": "value2"})
                     break
                 except Exception:
+                    # In the third attempt if get exception raise it
+                    if attempts == 3:
```
pmasrani: Please extract 3 as a constant/variable NUMBER_OF_ATTEMPTS instead of specifying it at multiple places..

packages/user_packages/iotcc_mqtt.py
```diff
@@ -97,6 +100,9 @@ def run(self, registry):
                     self.iotcc.set_properties(self.iotcc_edge_system, {"key1": "value1", "key2": "value2"})
                     break
                 except Exception:
```
pmasrani: Should we check for various exceptions? For example, if it is a version mismatch, it might not make sense to retry?

liota/dccs/iotcc.py
```diff
@@ -214,9 +219,29 @@ def create_relationship(self, reg_entity_parent, reg_entity_child):
...
+                        "result"] == "succeeded" and json_msg["body"]["parent"] == reg_entity_parent.reg_entity_id and \
+                                    json_msg["body"]["child"] == reg_entity_child.reg_entity_id:
+                        log.info(
+                            "Relationship between entities {0} & {1} created successfully in IoTCC".format(
+                                reg_entity_parent.ref_entity.name, reg_entity_child.ref_entity.name))
+                    else:
+                        log.info("Waiting for create relationship response")
+                        on_response(self.recv_msg_queue.get(True, timeout))
+                except Exception as err:
+                    log.exception("Exception while create relationship request")
```
pmasrani: Do we need to log as well as raise the err?

liota/dccs/iotcc.py
KohliDev: I have updated the pull request for each resource_registration, relationship, properties unregister request to have the separate queue for each call. By doing so we make it thread-safe. I have still kept the checks in on_response to be on safer side.

liota/dccs/iotcc.py
KohliDev: In the updated PR, transaction_id is now used for validating in "_dispatch_recvd_msg".

packages/user_packages/iotcc_mqtt.py
KohliDev: I think we need to discuss this as Liota already has unique topic generation based on uuid which was used in previous iotcc usecase.

packages/user_packages/iotcc_mqtt.py
KohliDev: Agreed, Check_version is recently introduced. I`m not sure how to handle it in user_code. Can you mark a separate issue for this change?

packages/user_packages/iotcc_mqtt.py
KohliDev: Fixed

KohliDev:
I have updated the pull request to handle multi-threaded user packages as per @GregBollella comment above. Changes are similar to Winnie code changes in v2 features branch.

Each request(registration, properties, relationship, unregister) call will have its own unique response_queue.
The dictonary of unique transaction_id and the response_queue will be maintained. The dictonary implementation in python is thred safe for single operations better to have locks which are introduced as part of this PR.

There will be thread blocking on the common respone queue part of MQTT callback. As the response is received on the basis of transaction_id they will be put into the respective queue.

KohliDev:
@pmasrani @GregBollella Please review the updated PR.

liota/dccs/iotcc.py
```diff
@@ -214,9 +238,35 @@ def create_relationship(self, reg_entity_parent, reg_entity_child):
             if entity_obj.unit is not None:
                 self.publish_unit(reg_entity_child, entity_obj.name, entity_obj.unit)
         else:
-            self.comms.send(json.dumps(self._relationship(self._next_id(),
+            # create relationship fun internal queue
```
pmasrani: fun?

config/liota.conf
```diff
@@ -10,6 +10,7 @@ entity_file_path = /usr/lib/liota/entity
 iotcc_path = /usr/lib/liota/iotcc.json
 enable_reboot_getprop = False
 iotcc_load_retry = 3
+iotcc_response_timeout = 300
```
pmasrani: Please add comment as to what the parameter does. Will be useful when someone wants to change it. Currently only liota developers understand these.

packages/user_packages/iotcc_mqtt.py
pmasrani: Agreed. Since this is in user example code, we can handle it out of this PR and even after code freeze date.

liota/dccs/iotcc.py
KohliDev: The log messages provide more details on where exactly the exception occurs(in which request) rather than looking at the stack trace.

liota/dccs/iotcc.py
KohliDev: The on_response() message is different for each request, in case of registration it returns uuid and in case of get_properties it is different as it returns the property list.  Also, there is specific check in the on_response call to process only the message which meets its entities criteria. Hence, better to have them separate as of now. Agreed in v2 we can surely refactor to have them common.

liota/dccs/iotcc.py
```diff
@@ -83,6 +90,12 @@ def __init__(self, con):
         self.enable_reboot_getprop = read_liota_config('IOTCC_PATH', 'enable_reboot_getprop')
         self.counter = 0
         self.recv_msg_queue = self.comms.userdata
+        self.req_ops_lock = Lock()
+        self.req_dict = {}
+        thread = threading.Thread(target=self._dispatch_recvd_msg)
```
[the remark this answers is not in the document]
KohliDev: Correctly pointed out, fixed.

liota/dccs/iotcc.py
pmasrani: For the existing on_response() methods in register and get_properties, the actual response gets processed to extract information. However, now we are introducing too many on_response methods  which can be handled by common code. Hence, it might be better to introduce new code with a common method, rather than duplicating it

liota/dccs/iotcc.py
```diff
@@ -102,31 +115,35 @@ def register(self, entity_obj):
...
                     log.exception("Exception while registering resource")
                     raise err
 
             if entity_obj.entity_type == "EdgeSystem":
                 entity_obj.entity_type = "HelixGateway"
+            transaction_id = self._next_id()
+            req = Request(transaction_id, reg_resp_q)
+            log.debug("Updating resource registration queue for transaction_id:{0}".format(transaction_id))
+            with self.req_ops_lock:
+                self.req_dict.update({transaction_id: req})
```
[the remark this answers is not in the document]
KohliDev: The implementation is kept similar to the changes in v2 where request object is used to hold Guid or can hold additional request parameters as well in future (alert_id in case of alert generation request). Hence, kept the implementation same.

liota/dccs/iotcc.py
```diff
@@ -245,7 +295,7 @@ def _relationship(self, msg_id, parent_res_uuid, child_res_uuid):
 
     def _properties(self, msg_id, entity_type, entity_id, entity_name, timestamp, properties):
         msg = {
-            "transationID": msg_id,
+            "transactionID": msg_id,
```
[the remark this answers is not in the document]
KohliDev: Agreed, more of refactoring of the pre-existing code can be taken up later.

liota/dccs/iotcc.py
```diff
@@ -314,11 +364,34 @@ def set_system_properties(self, reg_entity_obj, system_properties):
...
+                log.debug("Processing msg: {0}".format(json_msg["type"]))
+                self._check_version(json_msg)
+                if json_msg["type"] == "add_properties_response" and json_msg["body"][
+                    "result"] == "succeeded":
+                    log.info("System Properties defined for resource {0}".format(entity.name))
+                else:
+                    log.info("Waiting for set system properties response")
+                    on_response(set_sys_prop_resp_q.get(True, timeout), set_sys_prop_resp_q)
+            except Exception as err:
+                log.exception("Exception while setting system properties")
```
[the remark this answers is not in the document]
KohliDev: Log.exception to signify where the issue occurs rather than looking at stack trace.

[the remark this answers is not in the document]
pmasrani: It just prints the stracktrace twice unnecessarily. Log size increases as well as multiple stacktraces makes it more confusing. In fact, at some of the places, we are catching exception just to log and raise again.

liota/dccs/iotcc.py
```diff
@@ -713,10 +809,43 @@ def on_response(msg):
...
+            try:
+                # block until there is an item available
+                msg = self.recv_msg_queue.get(True)
+                log.debug("Received msg: {0}".format(msg))
+                json_msg = json.loads(msg)
+                log.debug("Processing msg: type:{0} transaction_id:{1}".format(json_msg["type"], json_msg["transactionID"]))
+                # search matched request in request dictionary
+                # assume transaction_id will be unique in one process
+                log.debug("Dictonary keys:{0}".format(self.req_dict.keys()))
+                if (json_msg["transactionID"] in self.req_dict):
```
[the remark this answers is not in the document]
KohliDev: req = self.req_dict[json_msg["transactionID"]] it will return keyerror not none hence the check has been introduced before.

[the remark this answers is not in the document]
pmasrani: Yes, you are right. Can we shift to self.req_dict.get(json_msg["transactionID"])?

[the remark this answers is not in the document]
KohliDev: That will be just refactoring?

liota/dccs/iotcc.py
KohliDev: Removed

liota/dccs/iotcc.py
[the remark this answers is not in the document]
KohliDev: Removed

liota/dccs/iotcc.py
```diff
@@ -713,10 +809,43 @@ def on_response(msg):
...
-
-        self.comms.send(json.dumps(self._get_properties(self._next_id(), resource_uuid)))
-        on_response(self.recv_msg_queue.get(True, timeout))
+        transaction_id = self._next_id()
+        req = Request(transaction_id, get_prop_resp_q)
+        log.debug("Updating get properties response queue for transaction_id:{0}".format(transaction_id))
+        with self.req_ops_lock:
+            self.req_dict.update({transaction_id: req})
+        self.comms.send(json.dumps(self._get_properties(transaction_id, resource_uuid)))
+        on_response(get_prop_resp_q.get(True, timeout),get_prop_resp_q)
```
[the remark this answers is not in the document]
KohliDev: Done

liota/dccs/iotcc.py
```diff
@@ -713,10 +809,43 @@ def on_response(msg):
...
         return self.prop_list
+
+    def _dispatch_recvd_msg(self):
+        log.info("Dispatching received messages from IOTCC")
```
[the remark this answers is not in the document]
KohliDev: Done

liota/dccs/iotcc.py
```diff
@@ -713,10 +809,43 @@ def on_response(msg):
...
+        while True:
+            try:
+                # block until there is an item available
+                msg = self.recv_msg_queue.get(True)
+                log.debug("Received msg: {0}".format(msg))
+                json_msg = json.loads(msg)
+                log.debug("Processing msg: type:{0} transaction_id:{1}".format(json_msg["type"], json_msg["transactionID"]))
+                # search matched request in request dictionary
+                # assume transaction_id will be unique in one process
+                log.debug("Dictonary keys:{0}".format(self.req_dict.keys()))
```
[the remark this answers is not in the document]
KohliDev: Done

liota/dccs/iotcc.py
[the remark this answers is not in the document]
pmasrani: Can we please not introduce new code to be refactored later? Not refactoring existing code right now should be fine and can be handled by raising bugs for them for future.

liota/dccs/iotcc.py
```diff
@@ -713,10 +809,43 @@ def on_response(msg):
...
+                    # get/delete request from dictionary
+                    if (req is not None):
+                        with self.req_ops_lock:
+                            del self.req_dict[json_msg["transactionID"]]
+                        # put response into requester's reception queue
+                        log.debug("Msg:{0} dispatched to user queue".format(msg))
+                        req.user_queue.put(msg)
+                else:
+                    # TBD: it may be other messages, e.g., Actions, Armada Campaign
+                    log.info("Received unexpected messages")
```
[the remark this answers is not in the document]
KohliDev: Done

liota/dccs/iotcc.py
```diff
@@ -713,10 +809,43 @@ def on_response(msg):
...
+            except Exception as er:
+                log.exception("Exception in dispatching received messages: %s" % str(er))
```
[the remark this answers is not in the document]
KohliDev: Done

config/liota.conf
KohliDev: Done

liota/dccs/iotcc.py
```diff
@@ -83,6 +90,12 @@ def __init__(self, con):
         self.enable_reboot_getprop = read_liota_config('IOTCC_PATH', 'enable_reboot_getprop')
         self.counter = 0
         self.recv_msg_queue = self.comms.userdata
+        self.req_ops_lock = Lock()
+        self.req_dict = {}
+        thread = threading.Thread(target=self._dispatch_recvd_msg)
+        thread.daemon = True
+        # This thread will continuously run in background to check and dispatch received responses
+        thread.start()
```
pmasrani: I was just wondering about the cleanup of this dcc. Once loaded, these threads keep on running forever. Can we do anything to stop them during cleanup step? Of course this applies to the existing thread as well along with this thread and can be handled separate from this pull request in future.

liota/dccs/iotcc.py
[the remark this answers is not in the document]
KohliDev: I was asking if the refactoring is the requirement with dict.get or some optimization? 

Fixed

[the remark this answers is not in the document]
pmasrani: Using dict.get() to remove the if...in check as we already have a check if returned object is not None

liota/dccs/iotcc.py
KohliDev: Agreed.

liota/dccs/iotcc.py
KohliDev: So we keep separate implementation for the request which needs to extract information on-response and separate refactoring for the call which just needs a successful check. Let`s discuss it on the call.

KohliDev: As discussed have refactored the code and have extracted out the processing of response messages for create_relationship, set_properties, set_system_properties and unregister requests. 

@pmasrani  Review the final code changes after incorporating all the review comments.

liota/dccs/iotcc.py
```diff
@@ -72,17 +81,23 @@ def __init__(self, con):
...
-        thread.start()
+        recv_thread.start()
         # Wait for Subscription to be complete and then proceed to publish message
         time.sleep(0.5)
         self._iotcc_json = self._create_iotcc_json()
         self._iotcc_json_load_retry = int(read_liota_config('IOTCC_PATH', 'iotcc_load_retry'))
         self.enable_reboot_getprop = read_liota_config('IOTCC_PATH', 'enable_reboot_getprop')
         self.counter = 0
         self.recv_msg_queue = self.comms.userdata
+        self.req_ops_lock = Lock()
```
pmasrani: Ideally, we should change all these variables to be internal only to this class (_req_ops_lock instead of req_ops_lock), but it will require changes to existing code as well. Hence, can be delayed to refactoring changes. Currently, if you want, can change the newly introduced one's to use _.

liota/dccs/iotcc.py
```diff
@@ -713,10 +783,43 @@ def on_response(msg):
...
+                # get/delete request from dictionary
+                if (req is not None):
+                    with self.req_ops_lock:
+                        del self.req_dict[json_msg["transactionID"]]
+                    # put response into requester's reception queue
+                    log.debug("Msg:{0} dispatched to user queue".format(msg))
+                    req.user_queue.put(msg)
+                else:
+                    # TBD: it may be other messages, e.g., Actions, Armada Campaign
+                    log.info("Received unexpected message {0}".format(msg))
```
pmasrani: Sorry for realizing this late. I think we can change this to ERROR or at least WARN, as we don't expect this to happen ever, and if it happens, probably something needs to be fixed.

liota/dccs/iotcc.py
KohliDev: Fixed. 

For others param will handle later.

liota/dccs/iotcc.py
```diff
@@ -214,12 +228,44 @@ def create_relationship(self, reg_entity_parent, reg_entity_child):
...
+        Process the responses for various types of request messages
+        :param msg: response message received
+        :return: boolean
+        """
+        try:
+            log.debug("Received msg: {0}".format(msg))
+            json_msg = json.loads(msg)
+            log.debug("Processing msg: {0}".format(json_msg["type"]))
+            self._check_version(json_msg)
+            if json_msg["body"]["result"] == "succeeded":
```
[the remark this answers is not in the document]
KohliDev: Fixed

liota/dccs/iotcc.py
KohliDev: Fixed to WARN

KohliDev:
@pmasrani I have incorporated the review comments. Hope PR can be approved for merging.

KohliDev:
Thanks. Merging the PR as I have some dependent code changes to handle.

PR #197, opened 2017-12-02
Changes to fix issues when package name contains space character

arunv10:
1. Changed message format written to the pipe from space char separated to comma char separated to allow space char in package names
2. Changed message read from the pipe to be split by comma char as delimiter
3. Used splitlines() instead of split() on the contents read from the packages_auto.txt to allow space char in package names

### DIFF
--- a/packages/liotad/liotapkg.sh
+++ b/packages/liotad/liotapkg.sh
@@ -95,9 +95,17 @@ then
     exit 6
 fi
 
+# Collect the supplied arguments and create a comma seperated string so that package manager can split using comma as
+# delimeter. The package name supplied could contain space character.
+comma_seperated_args="$1"
+shift
+for arg in "$@"; do
+	comma_seperated_args="$comma_seperated_args,$arg"
+done
+
 # Echo to named pipe
 #echo "Pipe file: $package_messenger_pipe" >&2
-echo "$@" > $package_messenger_pipe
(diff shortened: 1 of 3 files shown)


packages/liotad/liotapkg.sh
```diff
@@ -95,9 +95,17 @@ then
     exit 6
 fi
 
+# Collect the supplied arguments and create a comma seperated string so that package manager can split using comma as
+# delimeter. The package name supplied could contain space character.
+comma_seperated_args="$1"
+shift
+for arg in "$@"; do
```
pmasrani: i am not sure if i understand this. How does command line know start and stop of argument?

arunv10: $@ is an array containing command line arguments. The for loop in shell knows about start and end of an array.

Here, the first argument is copied and removed (using shift) from the beginning of the arguments array.  The remaining arguments are copied in a loop to created a comma separated string.

pmasrani: Understood. Thanks for the changes. As discussed, please add an example(s) in Readme as to how to execute the load command to take care of spaces. All other code changes seem fine.

arunv10: Thanks, Piyush. I have included a note in the README

arunv10:
@KohliDev @pmasrani  If changes are okay, can you please merge this PR?

PR #198, opened 2017-12-04
Avoid empty payload from liota

pmasrani:


liota/dccs/dcc.py
```diff
@@ -111,10 +111,11 @@ def publish(self, reg_metric):
             log.error("RegisteredMetric object is expected.")
             raise TypeError("RegisteredMetric object is expected.")
         message = self._format_data(reg_metric)
-        if hasattr(reg_metric, 'msg_attr'):
-            self.comms.send(message, reg_metric.msg_attr)
-        else:
-            self.comms.send(message, None)
+        if message:
```
arunv10: If there is no message, would it be better to raise an exception for the ABC implementers?

liota/entities/metrics/registered_metric.py
```diff
@@ -160,7 +160,7 @@ def send_data(self):
         """
         log.info("Publishing values for the resource {0} ".format(
             self.ref_entity.name))
-        if not self.values:
+        if self.values.qsize() == 0:
             # No values measured since last report_data
             return True
```
arunv10: Just an observation - send_data() returns True when there are no metrics queued. Otherwise, nothing is returned.
Should send_data() return True only if metric publish is successful and False, otherwise?

liota/dccs/dcc.py
pmasrani: This function is being called by send_thread() in liota_core, whenever a metric is supposed to be ready to send_data. It is not in the user thread, and hence, user will not be able to anything to catch this exception. Hence, it will not be good to throw this exception.

liota/entities/metrics/registered_metric.py
pmasrani: Very good point. There is no method consuming this return value for now. Since it is part of existing code, will not change it in this pull request and will defer for next changelist.

PR #199, opened 2017-12-05
Remove system properties explicit call

KohliDev:


config/liota.conf
```diff
@@ -17,6 +17,8 @@ enable_reboot_getprop = False
 iotcc_load_retry = 3
 # Configurable queue timeout in order to wait for the response messages from IOTCC
 iotcc_response_timeout = 300
+# System Properties List to be set during registration of Edge System/Devices
+system_properties= {}
```
pmasrani: We need to divide this into two groups or multiple files. We don't want users to set system_properties, but we want them to change iotcc_response_timeout.

liota/dccs/iotcc.py
```diff
@@ -90,6 +91,7 @@ def __init__(self, con):
         self._iotcc_json = self._create_iotcc_json()
         self._iotcc_json_load_retry = int(read_liota_config('IOTCC_PATH', 'iotcc_load_retry'))
         self.enable_reboot_getprop = read_liota_config('IOTCC_PATH', 'enable_reboot_getprop')
+        self.sys_properties = read_liota_config('IOTCC_PATH', 'system_properties')
```
pmasrani: Please change to _sys_properties

liota/dccs/iotcc.py
```diff
@@ -157,7 +159,15 @@ def on_response(msg, reg_resp_q):
                     self._store_reg_entity_attributes("Devices", entity_obj.name, self.reg_entity_id,
                                                       entity_obj.entity_type, None)
 
-            return RegisteredEntity(entity_obj, self, self.reg_entity_id)
+            _reg_entity_obj = RegisteredEntity(entity_obj, self, self.reg_entity_id)
+            if self.sys_properties is not None and self.sys_properties:
```
pmasrani: Aren't both conditions same?

liota/dccs/iotcc.py
```diff
@@ -157,7 +159,15 @@ def on_response(msg, reg_resp_q):
...
+                if ast.literal_eval(self.sys_properties):
```
pmasrani: What if this returns a String or a number of a tuple? I think we need a check if it is an instance of dictionary.

liota/dccs/iotcc.py
```diff
@@ -157,7 +159,15 @@ def on_response(msg, reg_resp_q):
...
+                    self.set_properties(_reg_entity_obj, ast.literal_eval(self.sys_properties))
+                    log.info(
+                        "System Properties defined for the resource {0}".format(entity_obj.name))
```
pmasrani: Can you please add the properties as well in the log? It will be useful to debug issues.

liota/dccs/iotcc.py
```diff
@@ -157,7 +159,15 @@ def on_response(msg, reg_resp_q):
                     self._store_reg_entity_attributes("Devices", entity_obj.name, self.reg_entity_id,
                                                       entity_obj.entity_type, None)
 
-            return RegisteredEntity(entity_obj, self, self.reg_entity_id)
+            _reg_entity_obj = RegisteredEntity(entity_obj, self, self.reg_entity_id)
+            if self.sys_properties is not None and self.sys_properties:
+                if ast.literal_eval(self.sys_properties):
+                    self.set_properties(_reg_entity_obj, ast.literal_eval(self.sys_properties))
```
pmasrani: What happens when set_properties() throws a ValueError? Ideally, it should not happen if ice-client is the only one setting it, but since system_properties is being defined in liota.conf, it can happen.

config/liota.conf
KohliDev: This can be taken up later in the current approach we can use the conf file to set the properties as it is required only by Airwatch which integrates with Liota as user.

pmasrani: Sure, we can take this up later. Please raise a bug for this. For the end users, we might end up telling them to change iotcc_response_timeout, collect_thread_pool_size, crl_path etc. but that should not lead them to try out changing other properties like system_properties, which are only for airwatch agent.

liota/dccs/iotcc.py
```diff
@@ -157,7 +159,18 @@ def on_response(msg, reg_resp_q):
                     self._store_reg_entity_attributes("Devices", entity_obj.name, self.reg_entity_id,
                                                       entity_obj.entity_type, None)
 
-            return RegisteredEntity(entity_obj, self, self.reg_entity_id)
+            _reg_entity_obj = RegisteredEntity(entity_obj, self, self.reg_entity_id)
+            if self._sys_properties:
+                _sys_prop_dict = ast.literal_eval(self._sys_properties)
+                if isinstance(_sys_prop_dict, dict) and _sys_prop_dict:
```
pmasrani: The second check is redundant as isinstance() takes care of None. Please remove it.

KohliDev: The second condition checks if the passed dictionary is not empty, I have tested it 'isinstance()' doesn't take care of that.

pmasrani: I tried out with the following code: 
#obj = {"a":"b"}
#obj = None
obj = {}
if isinstance(obj, dict) and obj:
  print "true"
else:
  print "false"

It printed true only when obj = {"a":"b"} and false for the other two. Am i missing something?

KohliDev: Yea because the condition is "if isinstance(obj, dict) and obj:", you have the second check which is part of the current PR as well.

pmasrani: Understood. Thanks for pointing it out.

PR #200, opened 2017-12-18
Revert "Merge pull request #191 from winniex1/get_property"

pmasrani:
This reverts commit 29b719895221b53f14aa09dd1c801667b0b6586c, reversing
changes made to fb2009dead31d938986b3da73f34e4106da7b3b1.

# Conflicts:
#	liota/lib/utilities/utility.py
#	packages/liotad/liotapkg.sh

### DIFF
--- a/liota/core/package_manager.py
+++ b/liota/core/package_manager.py
@@ -40,7 +40,8 @@ from threading import Thread, Lock
 from Queue import Queue
 from time import sleep
 from abc import ABCMeta, abstractmethod
-from liota.lib.utilities.utility import read_liota_config, sha1sum, validate_named_pipe
+
+from liota.lib.utilities.utility import read_liota_config, sha1sum
 
 log = logging.getLogger(__name__)
 
@@ -72,7 +73,6 @@ package_thread = None
 package_lock = None
 package_path = None
 package_messenger_pipe = None
-package_response_pipe = None
 
 # Parse Liota configuration file
 package_path = os.path.abspath(
@@ -81,12 +81,8 @@ package_path = os.path.abspath(
 package_messenger_pipe = os.path.abspath(
     read_liota_config('PKG_CFG', 'pkg_msg_pipe')
 )
-package_response_pipe = os.path.abspath(
-    read_liota_config('PKG_CFG', 'pkg_rsp_pipe')
-)
 assert(isinstance(package_path, basestring))
 assert(isinstance(package_messenger_pipe, basestring))
-assert(isinstance(package_response_pipe, basestring))
 
 package_startup_list_path = None
 package_startup_list = []
@@ -350,7 +346,6 @@ class PackageThread(Thread):
         self._packages_loaded = {}  # key: package name, value: PackageRecord obj
         self._resource_registry = ResourceRegistry()
         self._resource_registry.register("package_conf", package_path)
-        self._rsp_pipe_file = package_response_pipe
         self.flag_alive = True
         self.start()
 
@@ -482,17 +477,10 @@ class PackageThread(Thread):
 
             # Switch on message content (command), determine what to do
             command = msg[0]
-            if command in ["unload", "delete", "list", "stat", "unload_all",\
-                           "terminate"]:
-                # currently, write 'Success' to the named pipe once
-                #     successfully receives these commands
-                self._write_cmd_exec_status('Success\n')
-
-            # currently, write 'Success' to the named pipe after verification
-            #     of the checksum for load/reload/update related commands
             if command in ["load", "reload", "update"]:
                 #-----------------------------------------------------------
                 # Use these commands to handle package management tasks
+
                 with package_lock:
                     offset = 0
                     autoload_flag = False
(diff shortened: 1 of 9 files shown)


pmasrani:
Reply to Glen's comment: After reverting the changes, i tried out load, unload, unload_all, reload and list pkg commands.

liota/core/package_manager.py
```diff
@@ -40,7 +40,8 @@
 from Queue import Queue
 from time import sleep
 from abc import ABCMeta, abstractmethod
-from liota.lib.utilities.utility import read_liota_config, sha1sum, validate_named_pipe
+
```
KohliDev: Remove the extra line

liota/core/package_manager.py
```diff
@@ -482,17 +477,10 @@ def run(self):
...
-                # currently, write 'Success' to the named pipe once
-                #     successfully receives these commands
-                self._write_cmd_exec_status('Success\n')
-
-            # currently, write 'Success' to the named pipe after verification
-            #     of the checksum for load/reload/update related commands
             if command in ["load", "reload", "update"]:
                 #-----------------------------------------------------------
                 # Use these commands to handle package management tasks
+
```
KohliDev: Remove the extra line, reformat all the files using autopep format.

packages/liotad/liotapkg.sh
```diff
@@ -33,19 +33,18 @@
 
 liota_config="/etc/liota/liota.conf"
 package_messenger_pipe=""
-package_response_pipe=""
 
 if ! [ `ps -ef | grep "liotad.py" | grep -v grep | wc -l` -gt 0 ]
 then
         echo "Liota package manager is not running. Please run it first!"
-        exit 1
+        exit -1
```
KohliDev: What is the significance of changing it to "-1"????

I think they should be positive integers as they are independent of the rollback.

gkmccready: You're right. I missed it in my review. All exit values should be >= 0.

liota/core/package_manager.py
pmasrani: This is a revert of a commit. I don't want to introduce changes in a revert.

liota/core/package_manager.py
pmasrani: This is a revert of a commit. I don't want to introduce changes in a revert. Ideally, each developer should take care of formatting during initial check-in itself or there should be tests in the code base which fail if there are formatting issues.

packages/liotad/liotapkg.sh
pmasrani: This is a revert of a commit. I don't want to introduce changes in a revert. This is what happens when we address more than one issue in a single commit/merge :-)

I validated that ICE client is not consuming these values. It is still a bug in liota. Please raise a separate bug for this. Also, exit values 1 and 2 have special meaning, so we should not use them in the return status. Hence, with or without this revert, the bug still needs to be raised.

KohliDev: This is a new commit which is getting introduced and shouldn't change the fixed issue of exit status under the revert tag. It can be handled in separate commit but the correct code shouldn't be affected due to this commit. 

What is the significance of exit values "1" and "2" and "-1" and "-2"?

liota/core/package_manager.py
KohliDev: Now the changes are getting introduced in form of code revert and formatting should be done.

liota/core/package_manager.py
KohliDev: They are not changes, it is code formatting as this is getting introduced as part of the new commit.

liota/lib/utilities/utility.py
```diff
@@ -314,33 +314,42 @@ def read_user_config(config_file_path):
...
-    assert (isinstance(pipe_file, basestring))
-    if os.path.exists(pipe_file):
-        if stat.S_ISFIFO(os.stat(pipe_file).st_mode):
-            pass
+
+    def __init__(self):
+        pass
+
+    def validate_named_pipe(self, pipe_file):
+        """
```
KohliDev: I can see comment section added which wasn't part of Winnie`s initial commit. Hence, this is not a revert commit.

pmasrani: As discussed, this is a revert commit. utility.py and liotapkg.sh where having merge conflicts, and hence, these changes were manually merged.

A revert commit should only take care of reverting the changes, and should not introduce any new changes.

packages/liotad/liotapkg.sh
pmasrani: Exit code 1 and 2 have special meanings: http://tldp.org/LDP/abs/html/exitcodes.html

As discussed, this is a git revert commit, and we should not introduce new changes.

liota/core/package_manager.py
pmasrani: As discussed, formatting and all should not be taken care of as part of a git revert. It should ideally just revert whatever was introduced,

KohliDev:
The above-mentioned issues should be handled in separate pull request as it changes the existing code in the Liota codebase by the dev reverting the commit.

pmasrani:
Reply to Vaibhav: yes agreed. Please raise separate bugs for the issues. Can you please approve the changes if the revert seems fine?

KohliDev:
The issue has been opened and should be resolved, approving the pull request;

https://github.com/vmware/liota/issues/201

packages/liotad/liotapkg.sh
gkmccready: That page is describing what bash itself exits with; we can define whatever exit codes we want for when our scripts exit.

PR #202, opened 2017-12-21
Set SI units with prefix

KohliDev:
Set metric SI units with prefix

### DIFF
--- a/liota/dccs/iotcc.py
+++ b/liota/dccs/iotcc.py
@@ -407,6 +407,7 @@ class IotControlCenter(DataCenterComponent):
         """
          Publish SI units as properties for Metrics but RegisteredMetric object are simply returned
          so currently units are set to the parent RegisteredEntity(Device or EdgeSystem)
+         Either units are set with prefix as properties or only unit gets set as property if prefix doesn't exist.
         :param reg_entity_obj: RegisteredEntity Object
         :param metric_name: Metric Name
         :param unit: SI Unit
@@ -417,19 +418,20 @@ class IotControlCenter(DataCenterComponent):
             properties_unit_list = {
                 metric_name + "_unit": str_unit_name
             }
-            log.info("Publishing metric unit to IoTCC")
+            log.debug("Publishing unit {0} for metric {1} to IoTCC".format(str_unit_name, metric_name))
         elif isinstance(str_unit_name, basestring) and isinstance(str_prefix, basestring):
             properties_unit_list = {
                 metric_name + "_unit": str_unit_name,
                 metric_name + "_prefix": str_prefix
             }
-            log.info("Publishing metric unit with prefix to IoTCC")
+            log.debug("Publishing unit {0} with prefix {1} for metric {2} to IoTCC".format(str_unit_name, str_prefix,
+                                                                                           metric_name))
         else:
-            properties_unit_list = []
-            log.info("Metric unit with prefix cannot be parsed and published to IoTCC")
+            properties_unit_list = {}
+            log.debug("{0} metric unit with prefix cannot be parsed and published to IoTCC".format(metric_name))
         if properties_unit_list:
             self.set_properties(reg_entity_obj, properties_unit_list)
-            log.info("Published metric units")
+            log.info("Published units for metric {0}".format(metric_name))
 
     def _create_iotcc_json(self):
         msg = {
--- a/liota/dccs/iotcc.py
+++ b/liota/dccs/iotcc.py
@@ -418,20 +418,26 @@ class IotControlCenter(DataCenterComponent):
             properties_unit_dict = {
                 metric_name + "_unit": str_unit_name
             }
-            log.debug("Publishing unit {0} for metric {1} to IoTCC".format(str_unit_name, metric_name))
+            log.debug("Publishing unit {0} for metric {1} to IoTCC for resource {2}".format(str_unit_name, metric_name,
+                                                                                            reg_entity_obj.ref_entity.name))
         elif isinstance(str_unit_name, basestring) and isinstance(str_prefix, basestring):
             properties_unit_dict = {
                 metric_name + "_unit": str_unit_name,
                 metric_name + "_prefix": str_prefix
             }
-            log.debug("Publishing unit {0} with prefix {1} for metric {2} to IoTCC".format(str_unit_name, str_prefix,
-                                                                                           metric_name))
+            log.debug(
+                "Publishing unit {0} with prefix {1} for metric {2} to IoTCC for resource {3}".format(str_unit_name,
+                                                                                                      str_prefix,
+                                                                                                      metric_name,
+                                                                                                      reg_entity_obj.ref_entity.name))
         else:
             properties_unit_dict = {}
-            log.debug("{0} metric unit with prefix cannot be parsed and published to IoTCC".format(metric_name))
+            log.debug("{0} metric unit with prefix cannot be parsed and published to IoTCC for resource {1}".format(
+                metric_name, reg_entity_obj.ref_entity.name))
         if properties_unit_dict:
             self.set_properties(reg_entity_obj, properties_unit_dict)
-            log.info("Published units for metric {0}".format(metric_name))
+            log.info("Published units for metric {0} to IoTCC for resource {1}".format(metric_name,
+                                                                                       reg_entity_obj.ref_entity.name))
 
     def _create_iotcc_json(self):
         msg = {
--- a/liota/dccs/iotcc.py
+++ b/liota/dccs/iotcc.py
@@ -178,7 +178,10 @@ class IotControlCenter(DataCenterComponent):
 
     def _check_version(self, json_msg):
         if json_msg["version"] != self._version:
-            raise Exception("CLIENT SERVER VERSION MISMATCH. CLIENT VERSION IS: {0}. SERVER VERSION IS: {1}".format(self._version, json_msg["version"]))
+            raise Exception(
+                "CLIENT SERVER VERSION MISMATCH. CLIENT VERSION IS: {0}. SERVER VERSION IS: {1}".format(self._version,
+                                                                                                        json_msg[
+                                                                                                            "version"]))
 
     def unregister(self, entity_obj):
         """
@@ -287,12 +290,12 @@ class IotControlCenter(DataCenterComponent):
             "version": self._version,
             "type": "create_relationship_request",
             "body": {
-                "parent":  {
+                "parent": {
                     "kind": parent_entity.entity_type,
                     "id": parent_entity.entity_id,
                     "name": parent_entity.name
                 },
-                "child":  {
+                "child": {
                     "kind": child_entity.entity_type,
                     "id": child_entity.entity_id,
                     "name": child_entity.name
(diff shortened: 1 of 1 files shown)


KohliDev:
This PR fixes the issue which is introduced after 

https://github.com/vmware/liota/commit/3efd89c94bb0759e4fc547894f15e7521247683f

PR merge. 

If the metric SI unit(which doesn't contain prefix) is provided in the user package, the metric_name and prefix are not set.
"ValueError: The provided string contains unacceptable character :" gets thrown and package load fails.

KohliDev:
So it's either prefix && unit or just unit and never just prefix?
VK: Yes that`s correct.

 I've lost track of the schema for these metrics but since metrics need to be compared across MOs how are we managing the units if MOs choose the same metric but different units?
VK: This is a valid point, we can have deep dive discussion for v2 design phase.
The units are set as properties and user will have to manually check the units and compare the values. There is no query support in v1 to compare metrics published from two different MO`s.

gkmccready:
Okay, the first bit needs to be commented in the code.

The second bit... what should Liota be doing in this case? What is the property associated with? The MO or the metric? And that lack of query support is in the UI, but not the API, right? All I need to do is run two queries against the API and aggregate myself...

liota/dccs/iotcc.py
```diff
@@ -410,16 +413,23 @@ def publish_unit(self, reg_entity_obj, metric_name, unit):
...
-            str_prefix = ""
-        if not isinstance(str_unit_name, basestring):
-            str_unit_name = ""
-        properties_added = {
-            metric_name + "_unit": str_unit_name,
-            metric_name + "_prefix": str_prefix
-        }
-        self.set_properties(reg_entity_obj, properties_added)
-        log.info("Published metric unit with prefix to IoTCC")
+        if not isinstance(str_prefix, basestring) and isinstance(str_unit_name, basestring):
```
pmasrani: So the mentioned issue can happen only when SI units which do not have a prefix are used? No Error will happen if there is no SI unit is set, or if the SI unit is having a prefix. Do we have a list of units where the issue can happen and where it cannot?

liota/dccs/iotcc.py
```diff
@@ -410,16 +413,23 @@ def publish_unit(self, reg_entity_obj, metric_name, unit):
...
+            }
+            log.info("Publishing metric unit to IoTCC")
+        elif isinstance(str_unit_name, basestring) and isinstance(str_prefix, basestring):
+            properties_unit_list = {
+                metric_name + "_unit": str_unit_name,
+                metric_name + "_prefix": str_prefix
+            }
+            log.info("Publishing metric unit with prefix to IoTCC")
+        else:
+            properties_unit_list = []
```
pmasrani: Why is properties_unit_list a dictionary in if and elif but a list in else?

KohliDev: Good catch, I have corrected it.

liota/dccs/iotcc.py
```diff
@@ -410,16 +413,23 @@ def publish_unit(self, reg_entity_obj, metric_name, unit):
...
+            properties_unit_list = {
+                metric_name + "_unit": str_unit_name
+            }
+            log.info("Publishing metric unit to IoTCC")
+        elif isinstance(str_unit_name, basestring) and isinstance(str_prefix, basestring):
+            properties_unit_list = {
+                metric_name + "_unit": str_unit_name,
+                metric_name + "_prefix": str_prefix
+            }
+            log.info("Publishing metric unit with prefix to IoTCC")
```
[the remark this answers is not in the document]
KohliDev: Refactored the logs to debug and added units and prefix.

KohliDev:
@gkmccready 
I have added the comment.
The property is associated with MO hence metric_name_unit & metric_unit_prefix are added as property key.

Yes, it can be aggregated by running two queries against API, only UI lacks support in v1.

pmasrani:
Do we have a bug tracking this issue? Are we planning to check this in for RTM?

KohliDev:
QE doesn't have a bug tracking this issue, I strongly believe this should be checked-in as it doesn't affect the current Pulse UI but fixes the broken code.

pmasrani:
Any code change after RC should be avoided. Can you please get an approval from Arun and Prakash?

liota/dccs/iotcc.py
```diff
@@ -398,28 +401,37 @@ def set_properties(self, reg_entity_obj, properties):
...
-        if not isinstance(str_unit_name, basestring):
-            str_unit_name = ""
-        properties_added = {
-            metric_name + "_unit": str_unit_name,
-            metric_name + "_prefix": str_prefix
-        }
-        self.set_properties(reg_entity_obj, properties_added)
-        log.info("Published metric unit with prefix to IoTCC")
+        if not isinstance(str_prefix, basestring) and isinstance(str_unit_name, basestring):
+            properties_unit_list = {
```
pmasrani: Why is it called a list if it is a dictionary?

liota/dccs/iotcc.py
```diff
@@ -398,28 +401,37 @@ def set_properties(self, reg_entity_obj, properties):
...
+        elif isinstance(str_unit_name, basestring) and isinstance(str_prefix, basestring):
+            properties_unit_list = {
+                metric_name + "_unit": str_unit_name,
+                metric_name + "_prefix": str_prefix
+            }
+            log.debug("Publishing unit {0} with prefix {1} for metric {2} to IoTCC".format(str_unit_name, str_prefix,
+                                                                                           metric_name))
+        else:
+            properties_unit_list = {}
+            log.debug("{0} metric unit with prefix cannot be parsed and published to IoTCC".format(metric_name))
```
pmasrani: Is this the only situation when the code will reach else part? can it not reach here when no units are set?

I think the if..elif..else part is a bit confusing...can we change it to:

si_properties = {}
if (isinstance(str_unit_name, basestring):
          si_properties[metric_name + "_unit"]: str_unit_name
          if isinstance(str_prefix, basestring):
                   si_properties[metric_name + "_prefix"]: str_prefix
 if si_properties:
         self.set_properties()


What do you think?

liota/dccs/iotcc.py
```diff
@@ -398,28 +401,37 @@ def set_properties(self, reg_entity_obj, properties):
...
-            metric_name + "_unit": str_unit_name,
-            metric_name + "_prefix": str_prefix
-        }
-        self.set_properties(reg_entity_obj, properties_added)
-        log.info("Published metric unit with prefix to IoTCC")
+        if not isinstance(str_prefix, basestring) and isinstance(str_unit_name, basestring):
+            properties_unit_list = {
+                metric_name + "_unit": str_unit_name
+            }
+            log.debug("Publishing unit {0} for metric {1} to IoTCC".format(str_unit_name, metric_name))
```
pmasrani: It will also be good to get the entity name (and id?) in the log, as different entities can have the same metric name

liota/dccs/iotcc.py
```diff
@@ -398,28 +401,37 @@ def set_properties(self, reg_entity_obj, properties):
...
+                metric_name + "_prefix": str_prefix
+            }
+            log.debug("Publishing unit {0} with prefix {1} for metric {2} to IoTCC".format(str_unit_name, str_prefix,
+                                                                                           metric_name))
+        else:
+            properties_unit_list = {}
+            log.debug("{0} metric unit with prefix cannot be parsed and published to IoTCC".format(metric_name))
+        if properties_unit_list:
+            self.set_properties(reg_entity_obj, properties_unit_list)
+            log.info("Published units for metric {0}".format(metric_name))
```
pmasrani: entity name and id?

liota/dccs/iotcc.py
KohliDev: Entity_name gets printed in the set_properties call which is called from "publish_unit" method. Hence, I don't think it will be required to be added here.

liota/dccs/iotcc.py
KohliDev: Entity_name gets printed in the set_properties call which is called from "publish_unit" method. Hence, I don't think it will be required to be added here.

liota/dccs/iotcc.py
KohliDev: Thanks for the suggestion. But, I believe elif provides the clear explanation of all the three use cases and the better understanding of the three scenarios we can face with units and prefix.

liota/dccs/iotcc.py
KohliDev: Changed

liota/dccs/iotcc.py
pmasrani: As far as i understand, the else part will be visited even if no unit is set. The debug log in the else part does not indicate that. Please correct me if i am wrong.

liota/dccs/iotcc.py
pmasrani: It is not a good idea to keep related information in two different logs. Parsing multiple logs using automated parsers (like grep/logInsight) and coordinating them will be tough. Also, if set_properties call is modified in future, this single log will not provide all information and will have to be modified. So i still believe it will be good to add all information here, to make the log more useful

liota/dccs/iotcc.py
KohliDev: if entity_obj.unit is not None:                
self.publish_unit(reg_entity_child, entity_obj.name, entity_obj.unit)

If unit exists then only publish_unit method will be called.

pmasrani: ok, understood. Should publish_unit be an internal method (with an underscore) in that case?

KohliDev: yes it should be there are multiple methods which are required to be refactored in the similar way in the Liota codebase. And I believe that should be handled in separate PR as they are previously introduced methods by various developers.

pmasrani: I agree we should change it in whole code base later. However, iotcc.py is the primary module we will be exposing to liota package writers and hence, having public methods in this class will lead to unnecessary usage of these methods. I see only two methods in this class which are public but not part of the interface (get_properties and publish_unit: Hopefully i didnt miss any). Hence i think it will be better to at least make this method private. (get_properties might still be debatable and introduced as public knowingly.)

KohliDev: As, I mentioned above that entire change should be handled in separate PR and it should be marked as separate issue to revisit all methods of IoTCC. This PR is meant only for fixing the unit,prefix code changes.

pmasrani: Sure, I understand. Let's take it up later then.

liota/dccs/iotcc.py
KohliDev: Added entity_name as they remain unique for each entity and can be used to parse. All the log statements are currently not designed to be parsed using log insight by having a common identifier. I think all the log statements should be revisited in order to achieve that functionality.

KohliDev:
Addressed the review comments waiting for approval.

liota/dccs/iotcc.py
pmasrani: Sure, name is good as users can identify it quickly. Agreed about all the log statements not being in a very helpful state right now. I have asked the QE to raise bugs if they dont find some necessary information in logs. Also, as developers, we should continuously try to improve the existing logs, especially while introducing new ones. Thanks for addressing this one.

PR #203, opened 2017-12-21
Version bump in Liota for release

KohliDev:


KohliDev:
0.3.1 was in the plan to release. But there are bunch of changes which went in as part of v1.0 release cycles. Hence, bumping up the version to 0.4.

We have already have three pre-release versions:
https://github.com/vmware/liota/releases

Will mark 0.4 as latest release build for v1.0 at RTM.

setup.py
```diff
@@ -45,7 +45,7 @@
 #
 
 PACKAGE_NAME = "liota"
-PACKAGE_VERSION = "0.3.1"
+PACKAGE_VERSION = "0.4"
```
pmasrani: Having duplicate code and keeping on updating it in both places is not a good development practice. Keep only one setup.py and install liota-lite using an install parameter.

KohliDev: As per the guidance of John Hawley we have designed two setup.py for Liota and Liota-lite as both have different files to bundle in the Liota package. The feature request can be marked for the common setup.py and taken up in next release. Changing the approach now will require changes in Pulse installer before the RTM, which is not advised.

pmasrani: Sure, makes sense to delay it for now. Please raise a feature request to merge the two files.

KohliDev: Sure can you approve the PR?

PR #204, opened 2017-12-21
Default Value updated in Liota.conf 

KohliDev:
iotcc_response_timeout value updated to 600 as per test results.


KohliDev:
QE in there scale environment for 25000 MO`s (v1 support) have figured out that sometime responses takes time greater then 5 minutes but less then 10 minutes. Hence, setting the values with which testing has been done and verified.

PR #205, opened 2017-12-21
Readme updates

KohliDev:
CRL update and corrected Package manager readme formatting

PR #206, opened 2018-01-02
Readme Updates

KohliDev:
I have closed the previous pull request as it had merge conflict:
https://github.com/vmware/liota/pull/205

I'm raising this new pull request and have addressed the review comments from the previous PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,13 +36,14 @@ Liota supports plain old BSD sockets, WebSocket and MQTT communication protocols
 * Identity class encapsulates certificates or credentials related to a connection used at both DCC and Device side.
 * TLSConf class encapsulates parameters related to TLS configuration.
 * CRLs(Certificate Revocation Lists) check has been implemented: [VRIFY_CRL_CHECK_CHAIN](https://docs.python.org/2/library/ssl.html#ssl.VERIFY_CRL_CHECK_CHAIN) in this mode CRLs of all certificates in the peer cert chain are checked.
- In order to enable the CRLs set the 'crl_path' in '/etc/liota/liota.conf', it requires a valid path to CRL in PEM format (preferable directory '/usr/lib/liota'). If the check is required to be disabled 'crl_path' should be left blank.
+In order to enable the CRLs set the 'crl_path' in '/etc/liota/liota.conf', it requires a valid path to CRL; if in case intermediate CA is used the CRL should include both Intermediate CRL and Root CRL in PEM format (preferable directory '/usr/lib/liota'). If the check is required to be disabled 'crl_path' should be left blank.
+* Password based CA certificates are not supported.
 
 ## Package Manager
 Liota applications can be broken into small pieces that can be loaded and unloaded into a running liota process. We recommend putting the EdgeSystems, Devices, Metrics and DCC(s) in separate packages. Then each construct can be loaded and unloaded at will. See the README in the package directory for complete details.
 
 ## Device Discovery
-Liota SDK provides users a way to discover new devices at run-time through a dedicated discovery thread. The discovery thread could listen on a list of end points by spinning up a listening thread for each of them. Currently, Liota supports 4 kinds of end points: MQTT, COAP, Socket, and Named Pipe. However, because of security consideration on edge system, currently only MQTT and Named Pipe are allowed, especially Named Pipe is only accessible to its owner user.
+Under Development: Liota SDK provides users a way to discover new devices at run-time through a dedicated discovery thread. The discovery thread could listen on a list of end points by spinning up a listening thread for each of them. Currently, Liota supports 4 kinds of end points: MQTT, COAP, Socket, and Named Pipe. However, because of security consideration on edge system, currently only MQTT and Named Pipe are allowed, especially Named Pipe is only accessible to its owner user.
 
 Assume there is a list of DeviceType-to-DCC Mappings, such as {TypeA:[DCC1-Package-Name, DCC2-Package-Name], TypeB:[DCC3-Package-Name]}. Listening thread waits for a json message from devices, registers new devices with each DCC in the mapping list, sets any properties, fills out device file for AW agent. AW agent enrolls discovered device and pushes content to bring the device to compliance.
 
--- a/packages/liotad/README.md
+++ b/packages/liotad/README.md
@@ -13,12 +13,12 @@ Currently, supported commands include package action commands and statistical co
 
 ### Package action commands
 
-* **load** package_name <sha1_checksum> package_name_2 <sha1_checksum_package_2>
+* **load** package_name <sha1_checksum>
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -36,13 +36,14 @@ Liota supports plain old BSD sockets, WebSocket and MQTT communication protocols
 * Identity class encapsulates certificates or credentials related to a connection used at both DCC and Device side.
 * TLSConf class encapsulates parameters related to TLS configuration.
 * CRLs(Certificate Revocation Lists) check has been implemented: [VRIFY_CRL_CHECK_CHAIN](https://docs.python.org/2/library/ssl.html#ssl.VERIFY_CRL_CHECK_CHAIN) in this mode CRLs of all certificates in the peer cert chain are checked.
- In order to enable the CRLs set the 'crl_path' in '/etc/liota/liota.conf', it requires a valid path to CRL in PEM format (preferable directory '/usr/lib/liota'). If the check is required to be disabled 'crl_path' should be left blank.
+In order to enable the CRLs set the 'crl_path' in '/etc/liota/liota.conf', it requires a valid path to CRL; if in case intermediate CA is used the CRL should include both Intermediate CRL and Root CRL in PEM format (preferable directory '/usr/lib/liota'). If the check is required to be disabled 'crl_path' should be left blank.
```
pmasrani: In order to enable the CRLs set the 'crl_path' in '/etc/liota/liota.conf' --> In order to enable the CRLs set the 'crl_path' in liota.conf which by default is present in /etc/liota/.

pmasrani: Instead of saying "preferable directory", can we say, "for example /usr/lib/liota/"?

packages/liotad/README.md
```diff
@@ -1,43 +1,56 @@
...
 
-##PackageThread
+## PackageThread
 
 Currently, supported commands include package action commands and statistical commands.
 
-###Package action commands
+### Package action commands
+
+* **load** package_name <sha1_checksum> package_name_2 <sha1_checksum_package_2>
```
pmasrani: As "package_name_2 <sha1_checksum_package_2>" are not mandatory arguments, it will be good to write them in square brackets and mention they are optional. Here it feels that all these arguments are mandatory.

packages/liotad/README.md
```diff
@@ -1,43 +1,56 @@
...
 
-* **load** package_name sha1_checksum [package_name sha1_checksum] ...
+Load a package with the specified name and its sha1 checksum, even multiple packages can be loaded in the single command as shown above. For example, linux os user can first use "sha1sum filename" cmd to get checksum, and then load package:
```
pmasrani: ...checksum, even --> checksum. Even

packages/liotad/README.md
```diff
@@ -1,43 +1,56 @@
...
 
-Load a package with the specified name and its sha1 checksum. For example, linux os user can first use "sha1sum filename" cmd to get checksum, and then load package by
-"./liotapkg.sh load filename sha1_checksum".
+```bash
+$ ./liotapkg.sh load filename sha1_checksum
```
pmasrani: sha1_checksum --> <sha1_checksum> ?

packages/liotad/README.md
```diff
@@ -1,43 +1,56 @@
...
 ShouldUnregisterOnUnload = "True"
 ```
 
-* **reload** package_name sha1_checksum
+* **reload** package_name <sha1_checksum>
 
 Unload a package with the specified name and attempt to reload the same package **using the same file name**. Batch operation is not supported for reloading. If the specified package is not loaded when this command is invoked, the command will fail.
 
-* **update** package_name sha1_checksum [package_name sha1_checksum] ...
+* **update** package_name <sha1_checksum> ...
```
pmasrani: Can all the commands be mentioned with multiple packages? It might be good to mention which all can use multiple packages as parameters.

packages/liotad/README.md
```diff
@@ -48,19 +61,27 @@ Remove a package with the specified name. By default, the removed package will b
...
 There are 2 options to add liota package names and checksum:
+
 1. manually write into pkg_list file;
+
 2. add at run time through command [load, reload, update] by specifiying option of "-r", e.g.,
-"./liotapkg.sh load -r filename sha1_checksum".
+
+```bash
+$ cd /usr/lib/liota/packages
+$ ./liotapkg.sh load -r filename sha1_checksum
```
pmasrani: sha1_checksum --> <sha1_checksum> ?

README.md
KohliDev: I believe the current statement which is in the code review provides a better understanding.

KohliDev: I think preferable sounds better then "for example". As the other directories which user mentions is his responsbility.

packages/liotad/README.md
KohliDev: I have mentioned below "even multiple packages can be loaded in the single command as shown above". It clarifies that these arguments are not mandatory.

I think the multiple package load should only be mentioned in comments putting [optional], "..."will confuse users that it is part of the command line.

packages/liotad/README.md
KohliDev: Better to mention it in comments rather then showcasing in the command line.

packages/liotad/README.md
pmasrani: Ok, understood. Generally, the optional arguments are represented in square brackets. For example, in https://docs.docker.com/engine/reference/commandline/exec/
docker exec [OPTIONS] CONTAINER COMMAND [ARG...]
Similarly in https://kubernetes.io/docs/reference/kubectl/overview/
If you still think it is confusing, we can skip this.

packages/liotad/README.md
pmasrani: What i meant is, can all commands like unload, update, reload have multiple packages as arguments? It will good to mention in comments which all commands can take multiple packages as inputs and which one's not.

README.md
pmasrani: The current statement makes it sound mandatory to modify '/etc/liota/liota.conf' even if the actual location of liota.conf is different. My statement was just a suggestion. Its ok if you dont want to change it.

KohliDev: I plan to keep it in the current format.

packages/liotad/README.md
KohliDev: Plan to keep it in one sentence.

packages/liotad/README.md
KohliDev: changed

packages/liotad/README.md
KohliDev: Updated

packages/liotad/README.md
KohliDev: It has been added in in the comments.

KohliDev:
Addressed the review comments waiting for the approval.

packages/liotad/README.md
pmasrani: Personally, i don't think it is confusing. It definitely is not a standard, but a convention which is generally followed. Even linux man pages follow the same convention, so developers can easily understand it. We can address this later.

PR #210, opened 2018-01-05
User Package Update

KohliDev:
In this PR user packages are updated as per the test results and the retry mechanism for important calls like registration, create_relationship & set_properties. 

The PR needs to be merged urgently for the release. The retry logic values were finalized today in the meeting. Major review comments will be addressed.

### DIFF
--- a/packages/user_packages/README.md
+++ b/packages/user_packages/README.md
@@ -15,8 +15,9 @@ IoTCC over MQTT Protocol to acquire "registered edge system", i.e. iotcc_edge_sy
 which is pre-loaded during the installation in order to provide the required credentials and configuration parameters in the liota package manager registry.
 The properties for the edge system can be set as 'key:value' pair, you can also set the location by passing the
 'latitude:value' and 'longitude:value' as properties in the user package.
-Kindly include the unregister edge_system call in the clean up method required during the unload of the package. The retry mechanism has been implemented for important
-calls like registration, create_relationship or set_property in case of exception. User Configurable Retry and Delay Settings can be tweaked by user as per the targeted scale.
+If the unregister call is included in the clean up method then the resource will be unregistered and the entire history will be deleted
+from Pulse IoT Control Center so comment the logic if the unregsitration of the resource is not required to be done on the package unload.
+The retry mechanism has been implemented for important calls like registration, create_relationship or set_property in case of exception. User Configurable Retry and Delay Settings can be tweaked by user as per the targeted scale.
 
 * iotcc_mqtt_edge_system_stats.py
 
--- a/packages/user_packages/README.md
+++ b/packages/user_packages/README.md
@@ -26,7 +26,7 @@ Optional mechanism: If the device raises an intermittent exception during metric
 otherwise if an exception is thrown from user code the collection process will be stopped for that metric.
 If the None value is returned by User Defined Method(UDM) then metric value for that particular collector instance won't be published.
 
-* iotcc_mqtt_dev.py
+* iotcc_mqtt_device.py
 
 This is a sample device package which registers five devices to Pulse IoT Control Center and then a relationship is established to Edge System.
 A basic UDM returns random value it should be tweaked by user in order to collect device specific metric, all the five devices are loaded with dev_metric.
--- a/packages/user_packages/README.md
+++ b/packages/user_packages/README.md
@@ -15,7 +15,8 @@ IoTCC over MQTT Protocol to acquire "registered edge system", i.e. iotcc_edge_sy
 which is pre-loaded during the installation in order to provide the required credentials and configuration parameters in the liota package manager registry.
 The properties for the edge system can be set as 'key:value' pair, you can also set the location by passing the
 'latitude:value' and 'longitude:value' as properties in the user package.
-Kindly include the unregister edge_system call in the clean up method required during the unload of the package.
+Kindly include the unregister edge_system call in the clean up method required during the unload of the package. The retry mechanism has been implemented for important
+calls like registration, create_relationship or set_property in case of exception. User Configurable Retry and Delay Settings can be tweaked by user as per the targeted scale.
 
 * iotcc_mqtt_edge_system_stats.py
 
@@ -24,3 +25,10 @@ monitor the health status of the edge system from Pulse IoT Control Center.
 Optional mechanism: If the device raises an intermittent exception during metric collection process it will be required to be handled in the user code
 otherwise if an exception is thrown from user code the collection process will be stopped for that metric.
 If the None value is returned by User Defined Method(UDM) then metric value for that particular collector instance won't be published.
+
+* iotcc_mqtt_dev.py
+
+This is a sample device package which registers five devices to Pulse IoT Control Center and then a relationship is established to Edge System.
+A basic UDM returns random value it should be tweaked by user in order to collect device specific metric, all the five devices are loaded with dev_metric.
+The retry mechanism has been implemented for important calls like registration, create_relationship or set_property in case of exception.
+User Configurable Retry and Delay Settings can be tweaked by user as per the targeted scale.
--- a/packages/user_packages/iotcc_mqtt.py
+++ b/packages/user_packages/iotcc_mqtt.py
@@ -41,7 +41,8 @@ dependencies = ["general_edge_system", "credentials"]
 
 # --------------------User Configurable Retry and Delay Settings------------------------------#
 
-# The value mentioned below is in multiple of 1K, 2 for 2K, 3 for 3K, 4 for 4K and  5 for 5K Edge Systems
+# The value mentioned below is the total number of Edge System being deployed in the infrastructure
+# minimum 1 for 1K, 2 for 2K, 3 for 3K, 4 for 4K and  5 for 5K Edge Systems
 no_of_edge_system_in_thousands = 1
 # Number of Retries for Connection and Registrations
 no_of_retries_for_connection = 5
@@ -113,7 +114,7 @@ class PackageClass(LiotaPackage):
                                      client_id=registry.get("broker_username"), enable_authentication=True,
                                      mqtt_msg_attr=mqtt_msg_attr))
                     break
-                except Exception, e:
+                except Exception as e:
                     if conn_attempts == retry_attempts:
                         raise
                     conn_attempts += 1
@@ -128,11 +129,11 @@ class PackageClass(LiotaPackage):
                 try:
                     self.iotcc_edge_system = self.iotcc.register(edge_system)
                     break
-                except Exception, e:
+                except Exception as e:
                     if reg_attempts == retry_attempts:
                         raise
                     reg_attempts += 1
-                    log.error('MQTT Connection Failed - {0}'.format(str(e)))
+                    log.error('Exception while registering Edge System- {0}'.format(str(e)))
                     log.info('Trying Edge System {0} Registration: Attempt - {1}'.format(edge_system.name,
                                                                                          str(reg_attempts)))
                     time.sleep(delay_retries)
@@ -142,7 +143,8 @@ class PackageClass(LiotaPackage):
 
             # Attempts for setting edge system properties
             prop_attempts = 0
-            # Set Properties with retry attempts in case of exception
+            # Set multiple properties by passing Dictonary object for Edge System with the retry attempts
+            # in case of exceptions
             while prop_attempts < retry_attempts:
                 try:
                     self.iotcc.set_properties(self.iotcc_edge_system,
@@ -150,7 +152,7 @@ class PackageClass(LiotaPackage):
                                                "Location": "VMware HQ", "Building": "Promontory H Lab",
                                                "Floor": "First Floor"})
                     break
-                except Exception, e:
+                except Exception as e:
                     prop_attempts += 1
                     log.error(
                         'Exception while setting Property for Edge System {0} - {1}'.format(edge_system.name, str(e)))
@@ -159,12 +161,16 @@ class PackageClass(LiotaPackage):
                     time.sleep(delay_retries)
 
         except Exception:
-            log.error("EdgeSystem registration to IOTCC failed even after all the retries starting cleanup process")
-            self.clean_up()
+            log.error("EdgeSystem registration to IOTCC failed even after all the retries, starting connection cleanup")
+            # Disconnecting MQTT
+            self.iotcc.comms.client.disconnect()
             raise
 
     def clean_up(self):
         # Unregister the edge system
+        # On the unload of the package the Edge System will get unregistered and the entire history will be deleted
+        # from Pulse IoT Control Center so comment the below logic if the unregsitration of the device is not required
+        # to be done on the package unload
         self.iotcc.unregister(self.iotcc_edge_system)
         # Disconnecting MQTT
         self.iotcc.comms.client.disconnect()
--- a/packages/user_packages/iotcc_mqtt.py
+++ b/packages/user_packages/iotcc_mqtt.py
@@ -31,8 +31,35 @@
 # ----------------------------------------------------------------------------#
 
 from liota.core.package_manager import LiotaPackage
+import time
+import logging
+import random
+
+log = logging.getLogger(__name__)
 
 dependencies = ["general_edge_system", "credentials"]
+
+# --------------------User Configurable Retry and Delay Settings------------------------------#
+
+# The value mentioned below is in multiple of 1K, 2 for 2K, 3 for 3K, 4 for 4K and  5 for 5K Edge Systems
+no_of_edge_system_in_thousands = 1
+# Number of Retries for Connection and Registrations
+no_of_retries_for_connection = 5
+# MQTT Connection Delay Min Value in seconds
+mqtt_connection_retry_delay_min = 10
+# MQTT Connection Delay Max Value in seconds
+mqtt_connection_retry_delay_max = 600
+# Retry delay Min Value in seconds
+delay_retries_min = 600
+# Retry delay Max Value in seconds
+retries_delay_max = 1800
+
+# Lambda Function Multiplier uses the above settings for calculating retry and delay logic
+lfm = lambda x: x * no_of_edge_system_in_thousands
+retry_attempts = lfm(no_of_retries_for_connection)
+mqtt_connection_delay_retries = random.randint(lfm(mqtt_connection_retry_delay_min),
+                                               lfm(mqtt_connection_retry_delay_max))
+delay_retries = random.randint(lfm(delay_retries_min), lfm(retries_delay_max))
 
 
 class PackageClass(LiotaPackage):
@@ -44,19 +71,15 @@ class PackageClass(LiotaPackage):
     def run(self, registry):
         """
         The execution function of a liota package.
-
         Establishes connection with IoTControlCenter DCC using MqttDccComms
-
         :param registry: the instance of ResourceRegistryPerPackage of the package
         :return:
         """
         import copy
-        import time
         from liota.lib.utilities.identity import Identity
         from liota.dccs.iotcc import IotControlCenter
         from liota.dcc_comms.mqtt_dcc_comms import MqttDccComms
         from liota.lib.transports.mqtt import MqttMessagingAttributes
-        from liota.dccs.dcc import RegistrationFailure
         from liota.lib.utilities.tls_conf import TLSConf
 
         # Acquire resources from the registry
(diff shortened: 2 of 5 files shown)


KohliDev:
@swarno Please review the PR.

packages/user_packages/iotcc_mqtt.py
```diff
@@ -76,50 +99,73 @@ def run(self, registry):
...
-                    if attempts == max_retry_attempts:
+                except Exception, e:
+                    if reg_attempts == retry_attempts:
                         raise
-                    attempts += 1
-                    # The sleep time before re-trying depends on the infrastructure requirement of broker to restart
-                    # It can be modified or removed as per the infrastructure requirement
-                    time.sleep(5)
+                    reg_attempts += 1
+                    log.error('MQTT Connection Failed - {0}'.format(str(e)))
```
nareshbvemula: Change this to "Exception while GW Registration"

packages/user_packages/iotcc_mqtt_dev.py
```diff
@@ -0,0 +1,186 @@
+# -*- coding: utf-8 -*-
+# ----------------------------------------------------------------------------#
+#  Copyright © 2015-2016 VMware, Inc. All Rights Reserved.                    #
```
swarno: What is this dev file all about. Same as iotcc_mqtt.py?

swarno: Ok I see this is the child package that was being used and renamed. I don't understand the naming here though

packages/user_packages/iotcc_mqtt_dev.py
```diff
@@ -0,0 +1,186 @@
...
+                        self.iotcc.create_relationship(self.iotcc_edge_system, reg_dev)
+                        break
+                    except Exception, e:
+                        if relationship_attempts == retry_attempts:
+                            raise
+                        relationship_attempts += 1
+                        log.error(
+                            '{0} Device relationship with Edge System failed with following error - {1}'.format(
+                                device.name, str(e)))
+                        log.info(
```
swarno: Doesn't have the Trying log. Other logs are usually starting with Trying word.

packages/user_packages/iotcc_mqtt_dev.py
```diff
@@ -0,0 +1,186 @@
...
+                    try:
+                        reg_dev = self.iotcc.register(device)
+                        break
+                    except Exception, e:
+                        if reg_attempts == retry_attempts:
+                            raise
+                        reg_attempts += 1
+                        log.error(
+                            '{0} Device Registration Failed with following error - {1}'.format(device.name, str(e)))
+                        log.info('{0} Device Registration: Attempt: {1}'.format(device.name, str(reg_attempts)))
```
swarno: Doesn't have Trying word. Gateway registration messages had the word Trying.
This may help to find out similar logs with attempts.

packages/user_packages/iotcc_mqtt_dev.py
```diff
@@ -0,0 +1,186 @@
...
+                            '{0} Device relationship with Edge System failed with following error - {1}'.format(
+                                device.name, str(e)))
+                        log.info(
+                            '{0} Device Relationship: Attempt: {1}'.format(device.name, str(relationship_attempts)))
+                        time.sleep(delay_retries)
+                # Use the device name as identifier in the registry to easily refer the device in other packages
+                device_registry_name = socket.gethostname() + "-ChildDev" + str(i)
+                registry.register(device_registry_name, reg_dev)
+
+                # Attempts to set properties for registered device
```
swarno: Probably comment saying that it is trying to set multiple child properties by passing in a dictionary object would be more clear.

packages/user_packages/iotcc_mqtt.py
```diff
@@ -31,9 +31,36 @@
...
+import logging
+import random
+
+log = logging.getLogger(__name__)
 
 dependencies = ["general_edge_system", "credentials"]
 
+# --------------------User Configurable Retry and Delay Settings------------------------------#
+
+# The value mentioned below is in multiple of 1K, 2 for 2K, 3 for 3K, 4 for 4K and  5 for 5K Edge Systems
```
swarno: Add that this should be set based on the total number of gateways being deployed in your infrastructure.Minimum value should be 1. Should we also have a check for that?

packages/user_packages/iotcc_mqtt.py
```diff
@@ -76,50 +99,73 @@ def run(self, registry):
...
+
             registry.register("iotcc_mqtt", self.iotcc)
-            # Store the registered edge system object in liota package manager registry after the
             registry.register("iotcc_mqtt_edge_system", self.iotcc_edge_system)
 
-        except RegistrationFailure:
-            print "EdgeSystem registration to IOTCC failed"
+            # Attempts for setting edge system properties
+            prop_attempts = 0
+            # Set Properties with retry attempts in case of exception
```
swarno: Setting Multiple Properties of Gateway by passing Dictionary Object.

packages/user_packages/iotcc_mqtt.py
```diff
@@ -76,50 +99,73 @@ def run(self, registry):
...
-        The clean up function of a liota package.
-
-        Disconnects from IoTControlCenter DCC and un-registers the edge-system.
-
-        :return:
-        """
-        # Unregister the edge system on package unload
-        # Kindly include the edge system un-register call on package unload
+        # Unregister the edge system
         self.iotcc.unregister(self.iotcc_edge_system)
```
pmasrani: If somebody wants to unload the package for time being and has unregister call, it will end up deleting the history of the edge system/devices from Pulse IoT Center. Please add a comment as to when it is needed to have unregister() call and when it is not.

packages/user_packages/iotcc_mqtt_dev.py
swarno: If this is a new file should the Copyright be 2018-2019?

packages/user_packages/iotcc_mqtt.py
```diff
@@ -76,50 +99,73 @@ def run(self, registry):
...
+                    prop_attempts += 1
+                    log.error(
+                        'Exception while setting Property for Edge System {0} - {1}'.format(edge_system.name, str(e)))
+                    log.info('Trying setting properties for Edge System {0}: Attempt - {1}'.format(edge_system.name,
+                                                                                                   str(prop_attempts)))
+                    time.sleep(delay_retries)
+
+        except Exception:
+            log.error("EdgeSystem registration to IOTCC failed even after all the retries starting cleanup process")
+            self.clean_up()
```
pmasrani: If a gateway restarts after a few days, and tries properties registration and no response is received from Pulse, do we want to send unregister() edge system call to Pulse, deleting the whole history?

packages/user_packages/iotcc_mqtt.py
```diff
@@ -76,50 +99,73 @@ def run(self, registry):
...
+            while reg_attempts <= retry_attempts:
                 try:
-                    # Register edge system (gateway)
-                    self.iotcc.set_properties(self.iotcc_edge_system, {"key1": "value1", "key2": "value2"})
+                    self.iotcc_edge_system = self.iotcc.register(edge_system)
                     break
-                except Exception:
-                    # In the third attempt if get exception raise it
-                    if attempts == max_retry_attempts:
+                except Exception, e:
```
pmasrani: Do we want to retry for any Exception or only specific Exceptions?

packages/user_packages/iotcc_mqtt_dev.py
```diff
@@ -0,0 +1,186 @@
...
+                    self.metrics.append(reg_metric_simulated_received)
+
+                except Exception, e:
+                    log.error(
+                        'Exception while loading metric {0} for device {1} - {2}'.format(metric_name, device.name,
+                                                                                         str(e)))
+
+            except Exception:
+                log.info("Device Registration and Metrics loading failed starting the cleanup process")
+                self.clean_up()
```
pmasrani: For any Exception at all, do we want to call clean_up() and unregister all the devices completely?

packages/user_packages/iotcc_mqtt.py
```diff
@@ -31,9 +31,36 @@
...
+import random
+
+log = logging.getLogger(__name__)
 
 dependencies = ["general_edge_system", "credentials"]
 
+# --------------------User Configurable Retry and Delay Settings------------------------------#
+
+# The value mentioned below is in multiple of 1K, 2 for 2K, 3 for 3K, 4 for 4K and  5 for 5K Edge Systems
+no_of_edge_system_in_thousands = 1
```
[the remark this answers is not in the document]
KohliDev: It is as per the implementation done by @nareshbvemula & @swarno. They`ll be able to to provide more details.

packages/user_packages/iotcc_mqtt.py
KohliDev: Done

packages/user_packages/iotcc_mqtt.py
```diff
@@ -31,9 +31,36 @@
...
+mqtt_connection_retry_delay_max = 600
+# Retry delay Min Value in seconds
+delay_retries_min = 600
+# Retry delay Max Value in seconds
+retries_delay_max = 1800
+
+# Lambda Function Multiplier uses the above settings for calculating retry and delay logic
+lfm = lambda x: x * no_of_edge_system_in_thousands
+retry_attempts = lfm(no_of_retries_for_connection)
+mqtt_connection_delay_retries = random.randint(lfm(mqtt_connection_retry_delay_min),
```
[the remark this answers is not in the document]
KohliDev: mqtt_connection_delay_retries is used only once hence it is okay to be present at the top of the package.

delay_retries are used at multiple places but the calls are independent hence it should be okay to have them common. Also, the randomness is kind of estimation to avoid multiple packages retrying the same time from different edge_system which is addressed with the randomness provided by the above values.

packages/user_packages/iotcc_mqtt.py
```diff
@@ -76,50 +99,73 @@ def run(self, registry):
...
+            # Trying to establish MQTT Connection with retry attempts in case of exception
+            while conn_attempts <= retry_attempts:
+                try:
+                    self.iotcc = IotControlCenter(
+                        MqttDccComms(edge_system_name=edge_system.name, url=registry.get("broker_ip"),
+                                     port=registry.get("broker_port"), identity=identity, tls_conf=tls_conf,
+                                     client_id=registry.get("broker_username"), enable_authentication=True,
+                                     mqtt_msg_attr=mqtt_msg_attr))
+                    break
+                except Exception, e:
```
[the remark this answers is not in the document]
KohliDev: Fixed

packages/user_packages/iotcc_mqtt.py
KohliDev: Currently, the retry mechanism is handling all the exceptions.

packages/user_packages/iotcc_mqtt.py
KohliDev: Fixed

packages/user_packages/iotcc_mqtt.py
KohliDev: Added

packages/user_packages/iotcc_mqtt.py
KohliDev: The property call doesn't raise the exception after multiple attempt failure, hence cleanup of the entire history is not done.

But agreed we should remove the unregister call in case of multiple retry exception and should keep it only in the clean_up method with comments. I`ll fix it in both the packages.

packages/user_packages/iotcc_mqtt.py
KohliDev: Added

packages/user_packages/iotcc_mqtt_dev.py
KohliDev: This is the license file which was provided to us during the inception of Liota hence I`m not sure if it should be updated.

Renamed to iotcc_mqtt_device.py

packages/user_packages/iotcc_mqtt_dev.py
KohliDev: Added

packages/user_packages/iotcc_mqtt_dev.py
KohliDev: Added

packages/user_packages/iotcc_mqtt_dev.py
KohliDev: Added

packages/user_packages/iotcc_mqtt_dev.py
KohliDev: As mentioned above I have removed the clean_up call from both the packages.

packages/user_packages/README.md
```diff
@@ -15,7 +15,8 @@ IoTCC over MQTT Protocol to acquire "registered edge system", i.e. iotcc_edge_sy
 which is pre-loaded during the installation in order to provide the required credentials and configuration parameters in the liota package manager registry.
 The properties for the edge system can be set as 'key:value' pair, you can also set the location by passing the
 'latitude:value' and 'longitude:value' as properties in the user package.
-Kindly include the unregister edge_system call in the clean up method required during the unload of the package.
+Kindly include the unregister edge_system call in the clean up method required during the unload of the package. The retry mechanism has been implemented for important
```
pmasrani: Instead of saying "Kindly include the unregister call", for better understanding, we should say, "if unregister call is included in the cleanup method, the registered object can be unregistered and hence deleted from the server side."
 
I also noticed that you have provided a good comment in the package, so i think Readme should also reflect that. (You can reuse that)

packages/user_packages/iotcc_mqtt_edge_system_stats.py
```diff
@@ -127,62 +130,68 @@ def run(self, registry):
...
+            mem_free_metric = Metric(name=metric_name,
+                                     unit=None, interval=10,
+                                     aggregation_size=1,
+                                     sampling_function=read_mem_free
+                                     )
+            reg_mem_free_metric = iotcc.register(mem_free_metric)
+            iotcc.create_relationship(iotcc_edge_system, reg_mem_free_metric)
+            reg_mem_free_metric.start_collecting()
+            self.metrics.append(reg_mem_free_metric)
+        except Exception, e:
```
pmasrani: Please change here as well.

packages/user_packages/iotcc_mqtt_device.py
```diff
@@ -0,0 +1,190 @@
...
+        The clean up function of a liota package.
+        Unregister Device and Stops metric collection
+        :return:
+        """
+        # On the unload of the package the device will get unregistered and the entire history will be deleted
+        # from Pulse IoT Control Center so comment the below logic if the unregsitration of the device is not required
+        # to be done on the package unload
+        for device in self.reg_devices:
+            self.iotcc.unregister(device)
+            for metric in self.metrics:
```
pmasrani: Sorry for noticing this late, but should we not stop collecting metrics first and then unregister the device?

packages/user_packages/iotcc_mqtt_device.py
```diff
@@ -0,0 +1,190 @@
+# -*- coding: utf-8 -*-
+# ----------------------------------------------------------------------------#
+#  Copyright © 2015-2016 VMware, Inc. All Rights Reserved.                    #
```
pmasrani: I agree with Swarnadeep that generally in the copyright, current year is mentioned on creation of a new file. Up to you if you want to change it.

packages/user_packages/README.md
KohliDev: Added

packages/user_packages/iotcc_mqtt_device.py
KohliDev: We have been using this license comment at top of every Liota code file and if we change here it will require changes in the entire liota codebase, probably it should be reviewed from the OSL team before making this change, I`ll prefer to keep it same for now.

packages/user_packages/iotcc_mqtt_device.py
KohliDev: Changed

packages/user_packages/iotcc_mqtt_edge_system_stats.py
KohliDev: Changed

packages/user_packages/iotcc_mqtt_device.py
pmasrani: Now that you moved the unregister call later, i notice that all the metrics of all the devices are being collected in a single self.metrics, but while stopping the collection, it is being called for each device. Hence, both these loops should be separate. :-(

packages/user_packages/iotcc_mqtt_device.py
pmasrani: ok, no issues.

packages/user_packages/iotcc_mqtt_device.py
```diff
@@ -0,0 +1,190 @@
...
+                        prop_attempts = prop_attempts + 1
+                        log.error('Exception while setting property for Device {0} - {1}'.format(
+                            (device.name, str(e))))
+                        log.info('Trying setting properties for Device {0}: Attempt - {1}'.format(device.name,
+                                                                                                  str(prop_attempts)))
+                        time.sleep(delay_retries)
+
+                try:
+                    # Registering Metric for Device
+                    self.metrics = []
```
pmasrani: I had a discussion about this with @KohliDev. You were right, it seems self.metrics is getting re-initialized for every device and hence only the last metric will be present in self.metrics. So this also needs to be corrected.

packages/user_packages/iotcc_mqtt.py
```diff
@@ -76,50 +100,78 @@ def run(self, registry):
...
 
-        except RegistrationFailure:
-            print "EdgeSystem registration to IOTCC failed"
+            # Attempts for setting edge system properties
+            prop_attempts = 0
+            # Set multiple properties by passing Dictonary object for Edge System with the retry attempts
+            # in case of exceptions
+            while prop_attempts < retry_attempts:
+                try:
+                    self.iotcc.set_properties(self.iotcc_edge_system,
```
gkmccready: Hard coded strings about USA, California, Palo Alto, VMware, etc?

packages/user_packages/iotcc_mqtt_device.py
```diff
@@ -0,0 +1,190 @@
...
+
+log = logging.getLogger(__name__)
+
+dependencies = ["iotcc_mqtt"]
+
+# --------------------User Configurable Retry and Delay Settings------------------------------#
+
+# The value mentioned below is the total number of Edge System being deployed in the infrastructure
+# minimum 1 for 1K, 2 for 2K, 3 for 3K, 4 for 4K and  5 for 5K Edge Systems
+no_of_edge_system_in_thousands = 1
```
gkmccready: Should this be in a config file? Users shouldn't have to edit the source, right?

packages/user_packages/iotcc_mqtt_device.py
```diff
@@ -0,0 +1,190 @@
...
+                device_registry_name = socket.gethostname() + "-ChildDev" + str(i)
+                registry.register(device_registry_name, reg_device)
+
+                # Setting multiple properties by passing Dictonary object for Devices with the retry attempts
+                # in case of exceptions
+                prop_attempts = 0
+                while prop_attempts <= retry_attempts:
+                    try:
+                        self.iotcc.set_properties(reg_device,
+                                                  {"Country": "USA-G", "State": "California", "City": "Palo Alto",
```
gkmccready: Hard coded country, state, city, location, building, floor?

packages/user_packages/iotcc_mqtt_edge_system_stats.py
```diff
@@ -127,62 +130,68 @@ def run(self, registry):
...
-        mem_free_metric = Metric(name=metric_name,
-                                 unit=None, interval=10,
-                                 aggregation_size=1,
-                                 sampling_function=read_mem_free
-                                 )
-        reg_mem_free_metric = iotcc.register(mem_free_metric)
-        iotcc.create_relationship(iotcc_edge_system, reg_mem_free_metric)
-        reg_mem_free_metric.start_collecting()
-        self.metrics.append(reg_mem_free_metric)
+        try:
```
gkmccready: The changes in this file seem unrelated to the retry changes and should be a separate commit.

PR #216, opened 2018-02-16
set default to align with other params and so no failure from missing config value

dnance:
To align conndisconntimeout with the other params on the mqtt transport, initialize updating to set default rather than reading the config file by default.


KohliDev:
We want to show the usage of config value, hence the config file has been introduced.

PR #218, opened 2018-04-30
Fix liota package manager bug on incomplete command input

winniex1:


winniex1:
A bug was reported as "Liota : Liota package manager thread crashes on incomplete command to package manager". Please refer "https://jira-euc.eng.vmware.com/jira/browse/IOT-2706" for detailed explanation.
In a word, if you execute an Incomplete Command: sudo liotad/liotapkg.sh load, later no other commands could be executed. 
In the input, since only "load" command without any arguments (i.e, msg length is only 1), no msg[1] could be accessed, which caused a crash.
That's why I fix the issue. I don't want to change the code framework since it has been well tested. Just we have this bug of accessing char beyond range.

gkmccready:
Your fix is fine. But we should address the rest at some point, including the questions I asked.

winniex1:
Yes, will create an issue to memorize the rest.

winniex1:
Merge since Glen agreed the fix is fine for the bug.

PR #224, opened 2018-07-10
Integrate EdgeX

TiejunChina:
This is our preliminary effort to integrate EdgeX into Liota. Basically, EdgeX can build a device template specific to Liota via http server, and then load that device template to register that to Pulse IoT Center.

Signed-off-by: Huaqiao Zhang <huaqiaoz@vmware.com>


TiejunChina:
@KohliDev We resent out those two pull requests so just close this.  Sorry for this inconvenience this time.

